PR #100, opened 2026-04-19
Support Windows: TCP loopback fallback for AF_UNIX + UTF-8 stdio

bludragon66613-sys:
## Summary

Python's stdlib `socket` module on Windows (both python.org binaries and `uv`'s CPython builds) is compiled without `HAVE_UNIX_SOCKETS`, so `socket.AF_UNIX` raises `AttributeError` at import time — even on Windows 11 where the kernel itself supports AF_UNIX. Today that blocks the harness before it can reach Chrome.

This PR keeps the POSIX path untouched and adds a Windows fallback, gated on `hasattr(socket, "AF_UNIX")`.

### What changes

- **`daemon.py`**: when `AF_UNIX` is unavailable, bind `asyncio.start_server` to `127.0.0.1:0` (ephemeral loopback port) and write the chosen port to a sibling `.port` file in the same tempdir.
- **`helpers.py` / `admin.py`**: read the `.port` file and connect over `AF_INET` instead of `AF_UNIX`.
- **`admin.py` `restart_daemon`**: skip the `os.kill(pid, 0)` polling loop on Windows (`sig=0` raises `OSError` unconditionally there). The graceful `meta="shutdown"` message still runs on both platforms.
- **`run.py`**: `sys.stdout.reconfigure(encoding="utf-8")` so Windows' default cp1252 console doesn't `UnicodeEncodeError` on the `\U0001F7E2` tab-marker glyph in page titles.

POSIX behavior is unchanged — same socket path, same `chmod 0o600`, same pid-polling. The `tempfile.gettempdir()` call only runs on Windows; POSIX continues to hardcode `/tmp` for parity with existing daemons.

No new deps. `tempfile` is stdlib.

## Design notes

- Port file over random-but-stable port: avoids collisions when multiple `BU_NAME`s run concurrently, and still cleans up on shutdown (the `finally:` block in `serve()` unlinks it).
- `0o600` is POSIX-only, so Windows silently skips it. Loopback-only binding is the analogue; anything stronger would require Windows ACL manipulation.
- Kept the existing `/tmp/bu-<name>.sock` contract on POSIX so existing daemons/docs/skills keep working. Docs (`SKILL.md`, `interaction-skills/connection.md`) still mention `/tmp/bu-*.sock` — accurate on POSIX, wrong on Windows. Happy to sweep those in a follow-up if you prefer a single PR.

## Test plan

- [x] Windows 11 + Python 3.12 + Chrome 144 + `uv 0.11.6`: `new_tab`, `goto`, `wait_for_load`, `page_info`, `screenshot`, `list_tabs`, `restart_daemon` all work end-to-end against the user's real Chrome via `chrome://inspect/#remote-debugging`.
- [ ] macOS: not tested in this PR — code path is byte-identical to current `main` on POSIX. Would appreciate a reviewer spot-check.
- [ ] Linux: same as macOS.

---
## Summary by cubic
Add Windows support by falling back to TCP loopback when `socket.AF_UNIX` is missing, and switch stdio to UTF‑8 to prevent Windows console encoding errors. POSIX unix-socket behavior is unchanged.

- **New Features**
  - Windows: daemon binds `127.0.0.1:0` and writes the port to a `.port` file in `tempfile.gettempdir()`.
  - Clients (`admin.py`, `helpers.py`) read the `.port` file and connect over `AF_INET`.
  - `restart_daemon`: send graceful shutdown; skip `os.kill(pid, 0)` polling on Windows.

- **Bug Fixes**
  - Reconfigure `stdout`/`stderr` to UTF‑8 to avoid `UnicodeEncodeError` on Windows consoles.

<sup>Written for commit 9c0211e0f40d87919a7943c50f3291640a5908cd. Summary will update on new commits.</sup>

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -1,9 +1,15 @@
 import json
 import os
 import socket
+import tempfile
 import time
 import urllib.request
 from pathlib import Path
+
+# POSIX: AF_UNIX socket file. Windows: TCP loopback with port written to a
+# sibling .port file (stdlib Python on Windows is built without AF_UNIX).
+_USE_UNIX = hasattr(socket, "AF_UNIX")
+TMPDIR = "/tmp" if _USE_UNIX else tempfile.gettempdir()
 
 
 def _load_env():
@@ -26,11 +32,12 @@ BU_API = "https://api.browser-use.com/api/v3"
 
 def _paths(name):
     n = name or NAME
-    return f"/tmp/bu-{n}.sock", f"/tmp/bu-{n}.pid"
+    addr = os.path.join(TMPDIR, f"bu-{n}.sock" if _USE_UNIX else f"bu-{n}.port")
+    return addr, os.path.join(TMPDIR, f"bu-{n}.pid")
 
 
 def _log_tail(name):
-    p = f"/tmp/bu-{name or NAME}.log"
+    p = os.path.join(TMPDIR, f"bu-{name or NAME}.log")
     try:
         return Path(p).read_text().strip().splitlines()[-1]
     except (FileNotFoundError, IndexError):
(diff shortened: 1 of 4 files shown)


Alezander9:
We have implemented windows support now, thank you for raising this!

PR #104, opened 2026-04-19
Support Windows via cross-platform IPC (TCP loopback on Win, AF_UNIX unchanged on Unix)

rohitdutt108:
## Why

Browser Harness fails immediately at import time on Windows:

```
File "admin.py", line 42, in daemon_alive
    s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
```

Python on Windows does not expose `socket.AF_UNIX`, and `asyncio.start_unix_server` is not available either. The daemon IPC (one Unix socket at `/tmp/bu-<NAME>.sock` per daemon) is the only significant portability gap — everything else (CDP websocket, Chrome discovery, subprocess spawn) is already cross-platform.

## What this PR does

Introduces a thin `ipc.py` abstraction behind which:

- **Unix** keeps using `AF_UNIX` at `/tmp/bu-<NAME>.sock`, `asyncio.start_unix_server`, `chmod 600`, and `start_new_session=True` — **byte-identical to the current behavior**. Every Unix-specific line is preserved.
- **Windows** uses TCP loopback on an ephemeral port. The daemon writes its port number to `%TEMP%\bu-<NAME>.port`; the client reads it to connect. `asyncio.start_server` replaces `start_unix_server`. Subprocess spawn uses `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` instead of `start_new_session`.

Every platform branch is gated on `os.name == "nt"`, so Unix callers see zero behavioral change.

## Other small Windows fixes bundled

- `helpers.screenshot()` default path was `/tmp/shot.png`; now uses `tempfile.gettempdir()` on every platform.
- `admin.ensure_daemon` prefers `uv run daemon.py` when `uv` is on PATH, falling back to `[sys.executable, "daemon.py"]` when it is not — common on Windows where `uv tool install -e .` installs `browser-harness` globally but leaves `uv` itself in a project venv that is not on the system PATH.
- `run.py` calls `sys.stdout.reconfigure(encoding="utf-8")` at import. Harmless on Unix (already UTF-8). Necessary on Windows because the daemon prefixes the active tab's title with 🟢 (U+1F7E2), which the default cp1252 console cannot encode — `print(page_info())` raises `UnicodeEncodeError` otherwise.
- `pyproject.toml` — added `ipc` to `py-modules` so the editable install picks it up.

## Verified

Tested on Windows 11 + Python 3.13.9 against an isolated Chrome launched with `--remote-debugging-port=7341 --user-data-dir=C:\temp\bh-profile`:

- Daemon spawns detached and survives the parent's exit.
- `new_tab`, `wait_for_load`, `page_info`, `js`, `click`, `screenshot`, `cdp(...)` all work.
- `restart_daemon()` cleanly stops the daemon and unlinks the port/pid files.
- Daemon reuse on subsequent `browser-harness` invocations works (no respawn).
- No behavioral regression on helpers.py (other than the screenshot default path).

Unix behavior not retested by me — changes under `if os.name == "nt"` should leave Unix paths untouched; would appreciate a CI run or a quick local check on macOS/Linux before merge.

## Trade-offs to review

- TCP loopback on Windows means no `chmod 600` filesystem permission boundary. The port is bound to `127.0.0.1` so no external access, but any local user on the same Windows session could connect to the daemon if they discover the port. For the same-user-only threat model the upstream is designed for, I think this is acceptable — happy to tighten (e.g. token handshake) if reviewers disagree.
- Kept `uv run` as the preferred daemon invocation on Windows too (with fallback) to minimize behavioral drift. An alternative is to always use `sys.executable` — simpler but departs from the Unix convention. Preserved the existing invocation pattern instead.

## Commits

- `feat: support Windows via cross-platform IPC layer` — the core IPC abstraction.
- `fix(helpers): screenshot default path uses tempfile.gettempdir()` — one small follow-up.

Happy to squash/split on request.

---
## Summary by cubic
Add Windows support with a cross-platform IPC layer. Windows uses TCP loopback with a `.port` JSON file and per-daemon token; Unix keeps `AF_UNIX` sockets at `/tmp` unchanged.

- **New Features**
  - New `ipc` module abstracts IPC: Unix `/tmp/bu-<NAME>.sock`; Windows `127.0.0.1` with atomic `.port` JSON `{port, token}`. Provides `connect_client()`, `start_server()`, `listening_addr()`, `cleanup_addr()`.
  - Token auth on Windows and a `meta:"ping"` endpoint. Clients include the token; the daemon rejects unauthorized requests.
  - Detach daemon on Windows (`DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP`); Unix keeps `start_new_session=True`. Prefer `uv run daemon.py`, fallback to `[sys.executable, "daemon.py"]`. Added `ipc` to `py-modules`.

- **Bug Fixes**
  - Pin Unix socket path to `/tmp` to avoid macOS `AF_UNIX` path length issues.
  - Liveness checks require `{"pong": true}` to avoid false positives from stale `.port` files and port reuse.
  - Handle `TypeError` when parsing Windows `.port` JSON; invalid files now read as “no daemon running” instead of crashing.
  - `helpers.screenshot()` default path uses `tempfile.gettempdir()`; force UTF-8 on `sys.stdout`/`sys.stderr` to prevent `UnicodeEncodeError` on Windows.

<sup>Written for commit 81c014a9968b7847a44d268f7f87d717735d63a4. Summary will update on new commits.</sup>

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -34,12 +34,33 @@ def _log_tail(name):
 
 
 def daemon_alive(name=None):
+    # Connect-only isn't enough on Windows: if the daemon crashed and
+    # the OS reused its port for an unrelated listener, plain connect()
+    # would succeed against that other process. Send a ping with the
+    # token from the .port file and require the expected pong response
+    # so we confirm the listener is actually our daemon.
     try:
-        s = ipc.connect_client(name or NAME)
-        s.close()
-        return True
+        s, token = ipc.connect_client(name or NAME)
     except (FileNotFoundError, ConnectionRefusedError, socket.timeout, OSError):
         return False
+    try:
+        s.settimeout(1)
+        req = {"meta": "ping"}
+        if token:
+            req["token"] = token
+        s.sendall((json.dumps(req) + "\n").encode())
+        data = b""
+        while not data.endswith(b"\n"):
+            chunk = s.recv(1024)
+            if not chunk: break
+            data += chunk
+        r = json.loads(data or b"{}")
+        return r.get("pong") is True
+    except Exception:
+        return False
+    finally:
+        try: s.close()
+        except Exception: pass
 
 
 def ensure_daemon(wait=60.0, name=None, env=None):
@@ -102,8 +123,11 @@ def restart_daemon(name=None):
     n = name or NAME
     pidp = ipc.pid_path(n)
     try:
-        s = ipc.connect_client(n, timeout=5)
-        s.sendall(b'{"meta":"shutdown"}\n')
+        s, token = ipc.connect_client(n, timeout=5)
+        req = {"meta": "shutdown"}
+        if token:
+            req["token"] = token
+        s.sendall((json.dumps(req) + "\n").encode())
         s.recv(1024)
         s.close()
     except Exception:
--- a/admin.py
+++ b/admin.py
@@ -5,6 +5,8 @@ import time
 import urllib.request
 from pathlib import Path
 
+import ipc
+
 
 def _load_env():
     p = Path(__file__).parent / ".env"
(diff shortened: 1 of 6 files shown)


rohitdutt108:
Thanks @cubic-dev-ai — all three valid, pushed `2a9aaaa` to address them.

**P1 (ipc.py:17 — Unix path)** → Unix socket path pinned back at `/tmp/bu-<NAME>.sock` (and same for `.pid`, `.log`). `tempfile.gettempdir()` is now used only on Windows. Keeps Unix byte-identical to upstream and stays under the 108-char AF_UNIX path limit on macOS.

**P1 (daemon.py:208 — unauthenticated TCP)** → On Windows the daemon now generates a 32-byte hex token at startup and writes `{port, token}` JSON to the port file (atomic write-then-rename). `ipc.connect_client` returns `(socket, token)`; `Daemon.handle` rejects any request whose `req["token"]` doesn't match `ipc.expected_token()`. Unix keeps AF_UNIX + chmod 600 as the boundary — `_server_token` stays None there so the check is a no-op. Verified live: unauthorized connect → `{"error":"unauthorized"}`; correct token → `{"pong":true}`.

**P2 (daemon.py:222 — stale .port + port reuse)** → `already_running()` and `daemon_alive()` now do a proper `{"meta":"ping"}` handshake and require `{"pong":true}`. A stale port file pointing at an unrelated listener will fail this handshake and startup proceeds correctly.

`helpers._send` and `admin.restart_daemon` also forward the token on Windows. Full cold-start / reuse / unauthorized-reject / cleanup cycle verified on Windows 11.

rohitdutt108:
Good catch — pushed 81c014a. Now also catching TypeError so malformed `{port: null}` / array values fall through cleanly.

Alezander9:
Windows support was already added to the harness when I reviewed this, but I noticed some good additional pieces. I have lifted and merged them in https://github.com/browser-use/browser-harness/pull/276. Good work @rohitdutt108 ,+1vouch

PR #109, opened 2026-04-19
fix(helpers): dispatch_key sends wrong keycode for Delete/Home/End/Page*

MagicWand18:
## What

`dispatch_key(selector, key=X)` was falling back to `ord(X[0])` for any key not in the private `_KC` table. `_KC` only covered Enter, Tab, Escape, Backspace, Space, and the Arrow keys, so:

- `dispatch_key(sel, key="Delete")` → `keyCode 68` (ord of `D`) instead of `46`
- Same latent bug for `Home`, `End`, `PageUp`, `PageDown`

Listeners that check `e.keyCode` / `e.which` received the wrong value. Meanwhile `_KEYS` (used by `press_key`) already had the correct codes for all of these.

## Change

- Drop the `_KC` table.
- `dispatch_key` sources the keycode from `_KEYS`:

```python
kc = _KEYS[key][0] if key in _KEYS else (ord(key) if len(key) == 1 else 0)
```

Matches `press_key`'s semantics and removes the duplicate table. The keycode coverage of `dispatch_key` now grows automatically whenever `_KEYS` grows.

## Diff

`helpers.py`: +1 / -4. No new API, no behavior change for keys that were already correct.

---
## Summary by cubic
Fixes `dispatch_key` to read keyCode/which from `_KEYS` instead of a partial `_KC` table, so Delete, Home, End, PageUp, and PageDown emit the correct codes. Removes `_KC` and aligns behavior with `press_key`; no API changes.

<sup>Written for commit 3094c426dec59b4c719e0290a4e2c85a5343843d. Summary will update on new commits.</sup>

### DIFF
--- a/helpers.py
+++ b/helpers.py
@@ -184,16 +184,13 @@ def js(expression, target_id=None):
     return r.get("result", {}).get("value")
 
 
-_KC = {"Enter": 13, "Tab": 9, "Escape": 27, "Backspace": 8, " ": 32, "ArrowLeft": 37, "ArrowUp": 38, "ArrowRight": 39, "ArrowDown": 40}
-
-
 def dispatch_key(selector, key="Enter", event="keypress"):
     """Dispatch a DOM KeyboardEvent on the matched element.
 
     Use this when a site reacts to synthetic DOM key events on an element more reliably
     than to raw CDP input events.
     """
-    kc = _KC.get(key, ord(key) if len(key) == 1 else 0)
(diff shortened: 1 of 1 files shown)


MagicWand18:
Closing in favour of #825.

Same fix, rebased onto the current layout: this PR targeted `helpers.py` at the repo root, which moved to `src/browser_harness/helpers.py`, so it no longer applies cleanly.

The bug is still present on `main` — `_KC` (helpers.py:607) has no entry for `Delete`, `Home`, `End`, `PageUp` or `PageDown`, so `dispatch_key` emits `keyCode: 0` for all five.

PR #110, opened 2026-04-19
Windows support: AF_INET fallback for daemon IPC

unclesvf:
## Summary

Adds Windows support to Browser Harness. Currently the daemon and helpers unconditionally use `socket.AF_UNIX`, which Python on Windows does not expose — so every install on Windows fails immediately at `daemon.py` import / first IPC call:

```
```

(This is true even on Win10 1803+ where the OS has AF_UNIX support — Python's stdlib does not surface it.)

The repo also assumes `/tmp` exists, which Windows does not have.

## What changes

A new `_compat.py` module centralises the platform decision. POSIX paths are byte-for-byte unchanged; only Windows takes the new branch.

| Area | POSIX (unchanged) | Windows (new) |
|------|-------------------|---------------|
| Transport | AF_UNIX socket at `/tmp/bu-<NAME>.sock`, 0600 perms | AF_INET TCP on `127.0.0.1`, ephemeral port chosen by asyncio |
| Port discovery | (n/a — fixed socket path) | Daemon writes its bound port to `%TEMP%\bu-<NAME>.port`; clients read the file |
| PID/log | `/tmp/bu-<NAME>.{pid,log}` | `%TEMP%\bu-<NAME>.{pid,log}` (via `tempfile.gettempdir()`) |
| Subprocess detach | `start_new_session=True` | `CREATE_NEW_PROCESS_GROUP \| DETACHED_PROCESS` |
| Process-gone errors | `ProcessLookupError` | also `OSError` (Windows raises this when PID is gone) |

`admin.py`, `daemon.py`, and `helpers.py` are refactored to call the compat helpers (`client_connect`, `paths`, `start_server`, `remove_transport_artifacts`) instead of poking `AF_UNIX` directly. `pyproject.toml` lists `_compat` in `py-modules` so it ships with the editable install.

## Verified

Win11 + Python 3.13 (Anaconda) + uv:

```
$ uv run browser-harness <<'PY'
goto("https://goodmockups.com/4-free-fully-customizable-mug-mockup-psd-files/")
wait_for_load()
print(page_info())
PY
{'url': 'https://goodmockups.com/4-free-fully-customizable-mug-mockup-psd-files/',
 'title': '🟢 4 Free Fully Customizable Mug Mockup PSD Files - Good Mockups',
 'w': 500, 'h': 450, 'sx': 0, 'sy': 0, 'pw': 485, 'ph': 6358}
```

Daemon attaches to Edge via CDP, navigates, returns clean `page_info` (note the 🟢 prefix the harness adds to mark attached tabs).

## Risk to existing platforms

The POSIX code path is identical to before — only the syntactic indirection changes. The new `_compat.py` always returns the same socket type and same path on macOS/Linux that the inline code returned. CI on Linux/macOS should be byte-identical in behavior.

## Notes

- The `_compat.py` shim is intentionally small (~75 lines, no external deps beyond stdlib) so it stays easy to audit.
- Windows users still need to enable Chrome remote-debugging the same way POSIX users do (`chrome://inspect/#remote-debugging` → tick Discover → Allow). Edge equivalents work too.
- If you'd rather have the Windows branch live in `daemon.py` / `admin.py` directly with `if sys.platform == "win32"` blocks instead of a shim, happy to refactor — I picked the shim because the platform decision shows up in 6 places and inlining duplicates it.

---
## Summary by cubic
Adds Windows support for daemon IPC by falling back to AF_INET and moving artifacts to `%TEMP%`. Fixes the Windows import/connect crash while keeping macOS/Linux behavior unchanged.

- New Features
  - Added `/_compat.py` to choose transport per OS: AF_UNIX on POSIX, AF_INET on `127.0.0.1` (ephemeral port) on Windows.
  - Windows port discovery via `%TEMP%\bu-<NAME>.port`; PID/log also in `%TEMP%` (POSIX stays at `/tmp`).
  - POSIX sockets keep 0600 perms; Windows writes the chosen port to the sidecar file.

- Refactors
  - Switched `admin.py`, `daemon.py`, and `helpers.py` to `client_connect`, `paths`, `start_server`, and `remove_transport_artifacts`.
  - Windows daemon launch uses `CREATE_NEW_PROCESS_GROUP | DETACHED_PROCESS`; POSIX keeps `start_new_session=True`.
  - Gracefully handle dead PIDs on Windows by catching `OSError` in addition to `ProcessLookupError`.
  - Added `_compat` to `pyproject.toml` `py-modules` so it ships with the package.

<sup>Written for commit 77b3f85791429357cab64db49c6e0600d09fc89f. Summary will update on new commits.</sup>

### DIFF
--- a/_compat.py
+++ b/_compat.py
@@ -1,0 +1,89 @@
+"""Cross-platform IPC compat shim.
+
+POSIX (macOS/Linux) keeps AF_UNIX domain sockets — original behavior, unchanged.
+Windows uses AF_INET on 127.0.0.1 with a port published in a sidecar file
+because Python's socket module on Windows doesn't expose AF_UNIX (even though
+WSL-style Win10 1803+ supports it natively, the Python stdlib doesn't surface it).
+
+Daemon writes its bound port to <bu_dir>/bu-<name>.port at startup; clients
+read the file to find the port. PID/log files use the same per-name base.
+"""
+import os
+import socket
+import sys
+import tempfile
+from pathlib import Path
+
+IS_WINDOWS = sys.platform == "win32"
+
+
+def _bu_dir():
+    """Per-OS temp dir for daemon sockets/ports/pid/log files."""
+    if IS_WINDOWS:
+        return Path(tempfile.gettempdir())
+    return Path("/tmp")
+
+
+def paths(name):
+    """Return a dict of file paths for daemon `name`. Always has 'log' and 'pid'.
+    On POSIX also has 'sock'. On Windows also has 'port_file'."""
+    n = name or os.environ.get("BU_NAME", "default")
+    base = _bu_dir() / f"bu-{n}"
+    out = {"log": f"{base}.log", "pid": f"{base}.pid"}
+    if IS_WINDOWS:
+        out["port_file"] = f"{base}.port"
+    else:
+        out["sock"] = f"{base}.sock"
+    return out
+
+
+def client_connect(name=None, timeout=1.0):
+    """Connect a client socket to the daemon. Returns the connected socket.
+    Raises FileNotFoundError if the daemon isn't running, ConnectionRefusedError
+    or socket.timeout on transient failures (caller decides whether to retry)."""
+    p = paths(name)
+    if IS_WINDOWS:
+        try:
+            port = int(open(p["port_file"]).read().strip())
+        except (FileNotFoundError, ValueError):
+            raise FileNotFoundError(p["port_file"])
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.settimeout(timeout)
+        s.connect(("127.0.0.1", port))
+        return s
+    s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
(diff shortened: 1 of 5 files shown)


Qodo-Free-For-OSS:
Hi, On Windows, `_compat.client_connect()` reads the port discovery file via `open(...).read()` without closing the file, leaking a handle on every IPC call and eventually causing failures under repeated use.

**How to fix:** Use context manager for port

> ### Issue description
> On Windows, `_compat.client_connect()` opens the port-file without closing it, leaking a file handle on every IPC call.
> 
> ### Issue Context
> This is on a frequently executed path (`helpers._send()`), so the leak will accumulate quickly.
> 
> ### Fix Focus Areas
> - _compat.py[40-53]
> 
> ### Suggested change
> - Replace `open(...).read()` with `with open(...) as f: ...` (or `Path(...).read_text()`), and ensure sockets are also closed on connection exceptions if you touch that code.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review. FYI, Qodo is free for open-source.*

PR #119, opened 2026-04-20
fix: pin CDP session per-target to eliminate cross-tab drift

swapp1990:
## Summary

The daemon cached a single \`sessionId\` on \`self.session\` and reused it for every call. On Chrome M128+, where every visible tab has both a \`tab\` slot target and a \`page\` target underneath, the foreground-switch semantics plus BFCache / prerender swaps meant the cached session could legitimately point at a different page than the one we navigated — sometimes seconds after the user clicked another tab.

Fix: keep a \`{targetId → sessionId}\` map in the daemon, dispatch every call by \`targetId\` (auto-attach on first use), track \`current_tid\` for the default. This is how Puppeteer (\`CDPSession\` per Target) and Playwright (\`connectOverCDP\`) have done it from the start. One session per target, never pooled; the user switching tabs in Chrome's UI becomes a non-event.

## Repro (before this patch)

\`\`\`python
new_tab(\"https://example.com\")           # daemon attaches session S to tab A
# user clicks any other tab in Chrome
wait(10)
print(js(\"location.href\"))                # returns the other tab's URL, not example.com
\`\`\`

Looks like \"the daemon is rerouting my session,\" but it isn't — CDP is honoring the session ID. The session ID is just bound to a target whose page surface changed under it.

## What changed

- **\`daemon.py\`**: \`self.session\` → \`self.sessions: dict[targetId, sessionId]\` + \`self.current_tid\`. \`_attach()\` caches; \`_pick_real_page()\` finds or creates a real page.
- **\`is_real_page\`**: now also filters \`type == \"tab\"\` and \`subtype == \"prerender\"\` — both are new-ish and will mis-attach silently.
- **Event tap**: evicts from the map on \`Target.detachedFromTarget\` / \`Target.targetDestroyed\`; clears \`current_tid\` if the current target was destroyed.
- **Stale-target/stale-session**: one re-pick of a real page + retry, instead of a hard error.
- **WS handshake**: retries up to 30 s so the per-connection \"Allow DevTools\" dialog has time to be clicked. Previous 10 s budget made restarts flaky.
- **\`helpers.py\`**: \`cdp()\` and \`js()\` gain a \`target_id=\` kwarg that routes to the cached session for that target (no new session created). \`switch_tab()\` delegates the attach to the daemon via a \`set_target\` meta. \`current_tab()\` uses a \`current_target\` meta.

## Back-compat

External API is unchanged. Existing scripts that use \`cdp()\`, \`js()\`, \`switch_tab()\`, \`current_tab()\`, \`new_tab()\`, \`list_tabs()\` behave identically. Legacy \`session_id=\` passthrough still works.

## Tests (local)

| # | Test | Result |
|---|---|---|
| 1 | Smoke: \`new_tab\` → \`goto\` → \`js\` on one tab | pass |
| 2 | Explicit \`target_id\`: two tabs, \`js(expr, target_id=A)\` reads A without mutating current; subsequent \`js()\` still reads current (B) | pass |
| 3 | **Drift**: operator actively clicks to another Chrome tab; after 20 s the pinned session still returns the original URL | pass (confirmed by operator clicking) |
| 4 | Tab-close recovery: \`Target.closeTarget\` our attached tab, next \`js()\` auto-re-picks a real page and continues | pass |
| 5 | End-to-end: single bundled script scrapes notifications + For You feed + profile without any reattach workarounds | pass |

## Notes for reviewers

- **PR #115 (\`upload_file\` supports iframe \`target_id\`)** does per-call \`Target.attachToTarget\` + explicit \`session_id\`. With this change, it could simplify to \`cdp(..., target_id=target_id)\` and drop the per-call attach entirely (no session leak, no DOM-session juggling). Happy to coordinate.
- No changes to \`admin.py\`, \`run.py\`, docs, or skills.
- \`daemon.py\` diff is +157 / −51 but most of that is the \`handle()\` retry logic and new event tap branches. The core idea is the session map + \`_attach()\`.

## References

- [CDP Target domain](https://chromedevtools.github.io/devtools-protocol/tot/Target/) (the \`{type:\"browser\",exclude:true},{type:\"tab\",exclude:true}\` default filter note)
- [Puppeteer CDPSession](https://pptr.dev/api/puppeteer.cdpsession) (per-target session pattern)
- [Selenium issue #13300](https://github.com/SeleniumHQ/selenium/issues/13300) and [chromium-bidi #60](https://github.com/GoogleChromeLabs/chromium-bidi/issues/60): \"use sessionId instead of targetId to associate CDP client with a Context\" — same bug class, same conclusion.

## Test plan

- [ ] Reviewer: pull branch, run the repro script at top; session should stay pinned.
- [ ] Reviewer: run \`new_tab(\"https://example.com\")\`, \`close\` the tab manually, confirm \`js(\"location.href\")\` auto-recovers without raising.
- [ ] Reviewer: run the existing smoke tests used against top-of-main; nothing in the external API should have shifted.

---
## Summary by cubic
Pin a dedicated CDP session per `targetId` to stop cross‑tab drift on Chrome M128+. Commands now stay bound to the intended page even after tab switches, BFCache, or prerender swaps.

- **Bug Fixes**
  - Cache one session per target with `{targetId → sessionId}` and track `current_tid`; no pooling.
  - Attach only to real pages: ignore `type: "tab"` and `subtype: "prerender"`.
  - Evict on `Target.detachedFromTarget`/`Target.targetDestroyed`; on stale session/target, re-pick a page and retry once.
  - WebSocket handshake retries for up to 30s to allow the "Allow DevTools" prompt.

- **New Features**
  - `cdp()` and `js()` support `target_id=` to route a call to a specific tab without changing the current tab; reuses the cached session.
  - `switch_tab()` delegates attach to the daemon via `set_target`; `current_tab()` uses `current_target`.
  - External API unchanged; legacy `session_id=` still works.

<sup>Written for commit f24927f5ff80c23956ba06454723a0df819fb66f. Summary will update on new commits.</sup>

### DIFF
--- a/daemon.py
+++ b/daemon.py
@@ -97,50 +97,75 @@ def stop_remote():
 
 
 def is_real_page(t):
-    return t["type"] == "page" and not t.get("url", "").startswith(INTERNAL)
+    # "page" is a rendered frame; "tab" (M128+) is the slot container and must NOT be attached to.
+    # "prerender" page subtypes swap in/out and make sessions point at formerly-prerendered pages.
+    return (t["type"] == "page"
+            and t.get("subtype") != "prerender"
+            and not t.get("url", "").startswith(INTERNAL))
 
 
 class Daemon:
     def __init__(self):
         self.cdp = None
-        self.session = None
+        self.sessions = {}       # targetId -> sessionId
+        self.current_tid = None  # targetId of the "default" attached tab
         self.events = deque(maxlen=BUF)
         self.dialog = None
         self.stop = None  # asyncio.Event, set inside start()
 
-    async def attach_first_page(self):
-        """Attach to a real page (or any page). Sets self.session. Returns attached target or None."""
-        targets = (await self.cdp.send_raw("Target.getTargets"))["targetInfos"]
-        pages = [t for t in targets if is_real_page(t)]
-        if not pages:
-            # No real pages — create one instead of attaching to omnibox popup
-            tid = (await self.cdp.send_raw("Target.createTarget", {"url": "about:blank"}))["targetId"]
-            log(f"no real pages found, created about:blank ({tid})")
-            pages = [{"targetId": tid, "url": "about:blank", "type": "page"}]
-        self.session = (await self.cdp.send_raw(
-            "Target.attachToTarget", {"targetId": pages[0]["targetId"], "flatten": True}
+    async def _attach(self, target_id):
+        """Attach if not already; return sessionId. One session per target, cached."""
+        if target_id in self.sessions:
+            return self.sessions[target_id]
+        sid = (await self.cdp.send_raw(
+            "Target.attachToTarget", {"targetId": target_id, "flatten": True}
         ))["sessionId"]
-        log(f"attached {pages[0]['targetId']} ({pages[0].get('url','')[:80]}) session={self.session}")
+        self.sessions[target_id] = sid
         for d in ("Page", "DOM", "Runtime", "Network"):
             try:
                 await asyncio.wait_for(
-                    self.cdp.send_raw(f"{d}.enable", session_id=self.session),
+                    self.cdp.send_raw(f"{d}.enable", session_id=sid),
                     timeout=5
                 )
             except Exception as e:
-                log(f"enable {d}: {e}")
-        return pages[0]
+                log(f"enable {d} on {target_id}: {e}")
+        return sid
+
+    async def _pick_real_page(self):
+        """Find a real page target. Returns (targetId, url). Creates about:blank if none."""
+        targets = (await self.cdp.send_raw("Target.getTargets"))["targetInfos"]
+        pages = [t for t in targets if is_real_page(t)]
+        if pages:
+            return pages[0]["targetId"], pages[0].get("url", "")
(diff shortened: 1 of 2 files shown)


Alezander9:
Can you document more clearly what the issue is and if it is still there in todays version (things have changed a lot)?

Qodo-Free-For-OSS:
Hi, Daemon.start() recreates CDPClient repeatedly on handshake failures without explicitly closing the previous instance, which can leak resources if CDPClient.start() partially opens sockets before raising.

**Severity:** remediation recommended | **Category:** reliability

**How to fix:** Close prior client on retry

> ### Issue description
> On websocket handshake failures, the daemon retries by constructing a new `CDPClient` instance without cleaning up the previous one.
> 
> ### Issue Context
> Even if the typical failure happens before a socket is fully established, defensive cleanup avoids FD leaks in partial-connect scenarios.
> 
> ### Fix Focus Areas
> - daemon.py[152-161]
>   - Before reassigning `self.cdp` (or in the except block), attempt to close the previous client if it exists.
>   - If `CDPClient` exposes `close()`/`stop()`/`__aexit__`-style cleanup, use it; otherwise consider adding a small wrapper/try-finally pattern around `start()` attempts.

---

PR #129, opened 2026-04-20
feat: add Windows support via platform-switched IPC

ding-modding:
## Summary

Makes the harness run on Windows. The existing Unix code paths are preserved **byte-for-byte** on macOS/Linux; Windows branches are added only where the Unix primitive doesn't exist at all.

## Motivation

On Windows the harness fails immediately at `ensure_daemon()`:

```
```

Root causes, all cross-platform gaps in the current design:

| Unix primitive | Why it breaks on Windows |
|---|---|
| `AF_UNIX` + `/tmp/bu-*.sock` | `AF_UNIX` is absent from stock Windows Python; `/tmp` is not a path |
| `start_new_session=True` in `subprocess.Popen` | POSIX-only; silently ignored on Windows, so the daemon isn't detached |
| `os.kill(pid, 0)` for liveness | Any sig other than `CTRL_C_EVENT`/`CTRL_BREAK_EVENT` on Windows calls `TerminateProcess` — sending `0` kills the process |
| `os.kill(pid, SIGTERM)` for hard-kill fallback | Maps to `TerminateProcess` on Windows, so the daemon's `finally` (which `stop_remote()`s BU cloud browsers and unlinks state) never runs |

## Approach

Platform switch at each gap. Unix keeps its existing behavior exactly.

### IPC: AF_UNIX on Unix, TCP loopback on Windows

- **Unix** — unchanged. Still `AF_UNIX` at `/tmp/bu-{NAME}.sock`, still `os.chmod(0o600)`, still `asyncio.start_unix_server`.
- **Windows** — daemon binds `asyncio.start_server(host=\"127.0.0.1\", port=0)`, grabs the OS-assigned port, writes it to `{tempdir}/bu-{NAME}.port` atomically (`.tmp` + `os.replace`). Clients read the port file and connect over `AF_INET`.

### Path handling

Paths are resolved through a tiny `_tmp()` helper duplicated in each module (matching the existing `_load_env` duplication):

```python
def _tmp(name, suffix):
    if IS_WIN:
        return os.path.join(tempfile.gettempdir(), f\"bu-{name}.{suffix}\")
    return f\"/tmp/bu-{name}.{suffix}\"
```

Deliberately **not** using `tempfile.gettempdir()` on Unix. Two reasons:

1. On macOS, `gettempdir()` returns `/var/folders/xx/xx...xxx/T/` — per-user temp paths that can **exceed the 104-byte `AF_UNIX` limit** for long usernames.
2. On Linux, `$TMPDIR` would start overriding the fixed `/tmp/` path, which would be a silent behavior change for anyone who sets that env var.

Better to keep the Unix path literal.

### Subprocess detach

```python
if IS_WIN:
    kwargs[\"creationflags\"] = subprocess.CREATE_NEW_PROCESS_GROUP
else:
    kwargs[\"start_new_session\"] = True
```

`CREATE_NEW_PROCESS_GROUP` does two jobs: (1) detaches the daemon from the parent so it survives a fast `browser-harness` exit, and (2) makes the daemon the leader of its own process group, which we rely on for graceful shutdown below.

### `_pid_alive()` on Windows

```python
h = OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, False, pid)
if h: ...; return True
err = GetLastError()
return err != 87  # ERROR_INVALID_PARAMETER = pid truly gone
```

Important: **a NULL handle does not mean the pid is dead.** `ERROR_ACCESS_DENIED` (5) means the process is alive but in a different token. Treating that as dead would cause `restart_daemon()` to stop waiting early and unlink IPC state while the real daemon is still running.

### Graceful shutdown fallback

The existing flow is: send `{\"meta\":\"shutdown\"}` over the socket → daemon sets `stop.set()` → clean shutdown through the `finally` block (which `stop_remote()`s BU cloud browsers). If that hangs, we fall back to `os.kill(pid, SIGTERM)`.

On Windows, `SIGTERM` = `TerminateProcess` = `finally` never runs = **BU cloud browser keeps billing**. Patched to send `CTRL_BREAK_EVENT` instead — because we spawned the daemon with `CREATE_NEW_PROCESS_GROUP`, it's the group leader, receives the break as `KeyboardInterrupt`, and the `finally` block runs normally.

## Security trade-off

The Windows TCP loopback is **not ACL'd** the way the Unix `0600` UDS is. Any local process that guesses the port can send control messages to the daemon. On a single-user dev machine this blast radius is low — which is why I chose this over Windows named pipes (which would add meaningful complexity). It's called out in the source so future readers aren't surprised.

If stronger isolation is wanted later, adding a per-session shared-secret token in the port file closes the gap without new dependencies. Happy to do that in this PR if you want — flagging it as a choice, not an oversight.

## Testing

End-to-end on Windows 11 (Chrome 147, Python 3.13 via `uv`):

- ✅ Daemon spawn survives parent exit (`CREATE_NEW_PROCESS_GROUP` verified via `tasklist`)
- ✅ Port file atomic write (`bu-default.port` visible in `%TEMP%`)
- ✅ CDP WS attach via `BU_CDP_WS` override
- ✅ `new_tab` / `wait_for_load` / `page_info` / `click` / `screenshot` — all working against real sites (example.com, instagram.com including logged-in session)
- ✅ Unix paths: `/tmp/bu-*` literals unchanged in code; tested that `_tmp()` returns `/tmp/bu-...` on Linux and `/var/folders/.../bu-...` on... wait, no — `_tmp()` returns `/tmp/bu-...` on *both* Unix platforms now. That's the point.

## Out of scope (not fixed here)

- Chrome 147 on Windows sometimes fails to write `DevToolsActivePort` when launched with `--remote-debugging-port=9222` against the main user profile (listener never binds). Works fine with a fresh `--user-data-dir`. This is a Chrome-side quirk, not a harness issue. Users can work around by setting `BU_CDP_WS` directly or using a fresh profile. Worth a SKILL.md note eventually but out of scope here.
- Domain skills under `domain-skills/*/` still hardcode `/tmp/...` paths in example snippets. Unix users aren't affected; Windows users will hit these as they run each skill. Deferred to follow-up PRs so this one stays focused on the harness core.

## Review questions

1. OK with the Unix-paths-stay-literal decision (not `gettempdir()`)? Rationale is above.
2. OK with the TCP-loopback-no-auth trade-off for v1, or would you rather have token auth in this PR?
3. Any preference on docstring/SKILL.md updates, or keep them for a follow-up?

---
## Summary by cubic
Adds native Windows support by switching IPC to authenticated TCP loopback while leaving Unix `AF_UNIX` unchanged. Improves daemon startup/shutdown reliability and closes local IPC security gaps on Windows without impacting macOS/Linux.

- **New Features**
  - Windows IPC uses TCP on `127.0.0.1` with an ephemeral port and per-daemon token; both are written atomically to `{tempdir}/bu-{NAME}.port` and `bu-{NAME}.token`. Clients auto-read the token; liveness uses an authenticated ping.
  - Windows daemon spawns with `CREATE_NEW_PROCESS_GROUP`, checks liveness via `OpenProcess`, and stops via `CTRL_BREAK_EVENT`. Unix remains `AF_UNIX` at `/tmp/bu-{NAME}.sock` with `0600` perms and `start_new_session=True`.
  - `_tmp(...)` routes temp paths: Unix keeps `/tmp/bu-*.{sock,log,pid}`, Windows uses `tempfile.gettempdir()`. `screenshot(path=None, ...)` defaults to the system temp dir on Windows and `/tmp/shot.png` on Unix.

- **Bug Fixes**
  - Prevented false “daemon running” due to stale `.port` files by requiring an authenticated ping; `admin.daemon_alive()` now guards against non-dict JSON responses.
  - `restart_daemon()` waits for exit and only unlinks `PORT/PID/TOKEN` after confirming the pid is gone; uses `CTRL_BREAK_EVENT` on Windows instead of `SIGTERM` for graceful cleanup.
  - Admin and client helpers inject the token automatically on Windows, preserving existing call surfaces.

<sup>Written for commit 45922de4098d12144907f27b5a045c1449651b74. Summary will update on new commits.</sup>

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -87,7 +87,10 @@ def daemon_alive(name=None):
                 if not chunk: break
                 data += chunk
             s.close()
-            return bool(data) and json.loads(data).get("ok") is True
+            if not data:
+                return False
+            resp = json.loads(data)
+            return isinstance(resp, dict) and resp.get("ok") is True
         s.close()
         return True
     except (FileNotFoundError, ConnectionRefusedError, socket.timeout, ValueError, OSError, json.JSONDecodeError):
--- a/admin.py
+++ b/admin.py
@@ -44,6 +44,14 @@ def _paths(name):
     return _tmp(name, "port" if IS_WIN else "sock"), _tmp(name, "pid")
 
 
+def _token(name):
+    # Windows TCP loopback IPC requires a shared-secret token on every
+    # request. Missing/unreadable token => we can't talk to the daemon.
+    if not IS_WIN:
+        return None
+    return open(_tmp(name, "token")).read().strip()
+
+
 def _log_tail(name):
     try:
         return Path(_tmp(name, "log")).read_text().strip().splitlines()[-1]
@@ -65,10 +73,24 @@ def _connect(name, timeout=1):
 
 
 def daemon_alive(name=None):
+    # On Windows a stale PORT_FILE can point at a port reused by an
+    # unrelated process that happily accepts TCP connects. Send an
+    # authenticated ping and require the token-checked {"ok":true}
+    # reply before calling the daemon alive.
     try:
-        _connect(name, timeout=1).close()
+        s = _connect(name, timeout=1)
+        if IS_WIN:
+            s.sendall((json.dumps({"meta": "ping", "token": _token(name)}) + "\n").encode())
+            data = b""
+            while not data.endswith(b"\n"):
+                chunk = s.recv(4096)
+                if not chunk: break
+                data += chunk
+            s.close()
+            return bool(data) and json.loads(data).get("ok") is True
+        s.close()
         return True
-    except (FileNotFoundError, ConnectionRefusedError, socket.timeout, ValueError, OSError):
+    except (FileNotFoundError, ConnectionRefusedError, socket.timeout, ValueError, OSError, json.JSONDecodeError):
         return False
 
 
@@ -153,9 +175,16 @@ def restart_daemon(name=None):
     import signal
 
     ipc_path, pid_path = _paths(name)
+    token_path = _tmp(name, "token") if IS_WIN else None
     try:
         s = _connect(name, timeout=5)
-        s.sendall(b'{"meta":"shutdown"}\n')
+        shutdown_req = {"meta": "shutdown"}
+        if IS_WIN:
+            try:
+                shutdown_req["token"] = _token(name)
+            except (FileNotFoundError, OSError):
+                pass
+        s.sendall((json.dumps(shutdown_req) + "\n").encode())
         s.recv(1024)
         s.close()
     except Exception:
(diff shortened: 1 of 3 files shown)


Alezander9:

Issue #124, opened 2026-04-20
Windows support: 4 issues found during fresh install (AF_UNIX, /tmp paths, emoji encoding, Chrome 147 user-data-dir)


seltzdesign:
Tested on: Windows 11 Pro 26200, Chrome 147.0.7727.102, Python 3.14 (uv-managed)

Four issues hit during a fresh install on Windows. Fixes for all four are straightforward.
Note: websockets 16.0 + Chrome 147 CDP failure is already tracked in #106.

---

## 1. `socket.AF_UNIX` not available in uv-managed Python builds

All Python versions installed via `uv python install` on Windows lack
`socket.AF_UNIX`, even though Windows 10+ (build 17093+) supports it at the OS
level. The python-build-standalone distributions that uv uses are compiled
without it.

**Error:**
```
AttributeError: module 'socket' has no attribute 'AF_UNIX'
```

This crashes `daemon_alive()` immediately on startup before any connection
attempt. Same root cause kills `asyncio.start_unix_server` (not defined on
Windows either).

**Fix:** TCP fallback on Windows. In `admin.py`/`daemon.py`/`helpers.py`,
detect `platform.system() == "Windows"` and use `asyncio.start_server` on
`127.0.0.1:0`, writing the assigned port to a `.port` file alongside the
existing `.pid` file.

---

## 2. Hardcoded `/tmp/` paths don't exist on Windows

`daemon.py` and `helpers.py` hardcode `/tmp/bu-{NAME}.sock` etc.
`/tmp/` doesn't exist on Windows — `tempfile.gettempdir()` returns
`C:\Users\<user>\AppData\Local\Temp`.

**Fix:** Replace the `/tmp/` prefix with `tempfile.gettempdir()` in both files.

---

## 3. Chrome 147 silently refuses `--remote-debugging-port` with the default user data directory

When Chrome is launched with `--remote-debugging-port=9222` but without an
explicit `--user-data-dir`, Chrome 147 prints the following to stderr and never
creates `DevToolsActivePort` or opens any TCP port:

```
DevTools remote debugging requires a non-default data directory.
Specify this using --user-data-dir.
```

Chrome starts and opens normally — there is no visible error to the user, and
no hint in the daemon's error output either. The daemon just times out after
60 s with "DevToolsActivePort not found".

**Workaround:** always launch Chrome with an explicit `--user-data-dir` pointing
to a non-default path, e.g.:

```
[...]

Issue #106, opened 2026-04-19
[P0] cdp_use 1.4.5 + websockets 16.0 + Chrome 147: CDP WebSocket fails immediately (Sec-WebSocket-Extensions header)


ythx-101:
## Summary

With `cdp_use==1.4.5` (which depends on `websockets>=16.0`) and Chrome 147, the daemon cannot establish a CDP WebSocket connection. `browser-harness` exits immediately after install with an EOF error.

## Steps to Reproduce

```bash
uv tool install -e .   # installs cdp_use 1.4.5 + websockets 16.0
/usr/bin/chromium --headless --no-sandbox --remote-debugging-port=9222 --disable-gpu &
sleep 2
browser-harness <<< "print(page_info())"
```

## Error

```
fatal: CDP WS handshake failed — EOF after 7 bytes
```

## Root Cause

`websockets 16.0` sends an extra header during WebSocket upgrade:
```
Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits
```

Chrome 147's CDP endpoint rejects this header and immediately closes the connection (sends ~7 bytes then EOF). Without this header (e.g. raw socket test), Chrome responds with HTTP 101 normally.

**Verified via manual socket test:**
```python
import socket
s = socket.socket()
s.connect(("127.0.0.1", 9222))
request = "GET /devtools/browser/... HTTP/1.1\r\nHost: 127.0.0.1:9222\r\nUpgrade: websocket\r\nConnection: Upgrade\r\nSec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==\r\nSec-WebSocket-Version: 13\r\n\r\n"
s.sendall(request.encode())
print(s.recv(1024))  # → HTTP/1.1 101 WebSocket Handshake ✅
```
Adding `Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits` → EOF ❌

## Version Matrix

| websockets | Chrome | Result |
|------------|--------|--------|
| 16.0       | 147    | ❌ EOF after 7 bytes |
| ≤15.x      | 147    | ✅ (expected) |
| 16.0       | <144   | ✅ (expected) |

## Environment

- OS: Linux aarch64 (container)
- Python: 3.11
- Chromium: 147.0.7727.55
- cdp_use: 1.4.5
- websockets: 16.0
- browser-harness: main @ 2026-04-19

## Suggested Fixes

1. **Pin `websockets<16` in cdp_use** (fastest mitigation)
[...]

PR #142, opened 2026-04-21
Fix Chrome 147+ blocking remote debugging on default profile (Linux, macOS, Windows)

### LINKED ISSUE #107: [UX] Chrome remote debugging is a hidden prerequisite — not mentioned in README Quick Start
## Summary

The README Quick Start does not mention that Chrome must be launched with remote debugging enabled before running `browser-harness`. First-time users encounter a confusing error that only hints at the requirement indirectly.

## Current Flow

1. User follows README: installs uv → clones repo → `uv tool install -e .`
2. Runs `browser-harness` → error listing Chrome paths → must infer that remote debugging is needed
3. Searches `chrome://inspect/#remote-debugging` — not mentioned in Quick Start

The only mention of "enable remote debugging" is inside a screenshot in the **setup prompt** section, not Quick Start.

## Suggested Fix

Add to Quick Start:

```markdown
3. Launch Chrome with remote debugging enabled:
   # macOS
   open -a 'Google Chrome' --args --remote-debugging-port=9222
   # Linux
   google-chrome --remote-debugging-port=9222

   Then visit chrome://inspect/#remote-debugging and enable "Discover network targets"
```

## Notes

- The error message itself is **good** — it lists which Chrome paths were checked and mentions `chrome://inspect`. But Quick Start users shouldn't need an error to find a prerequisite.
- `uv` is also an unlisted prerequisite (separate issue worth filing).

---
*Reported by [test-factory](https://github.com/ythx-101/test-factory) walk-through audit, 2026-04-20*

### LINKED ISSUE #123: README Quick Start: add Chrome remote debugging prerequisite
## Problem

The README Quick Start section does not mention that Chrome must be launched with `--remote-debugging-port` before running browser-harness. First-time users hit a confusing error and must infer the requirement from error messages.

## Solution

Add a step to the Quick Start section before the run command:

```
3. Launch Chrome with remote debugging enabled:

   # macOS
   open -a 'Google Chrome' --args --remote-debugging-port=9222

   # Linux
   google-chrome --remote-debugging-port=9222

   Then visit chrome://inspect/#remote-debugging and enable "Discover network targets"
```

The uv prerequisite is a separate issue; this focuses only on the Chrome prerequisite gap.

## Use case

New users following the Quick Start who encounter the "no Chrome found" error before they even get started.

1RB:
## Summary

Chrome 147+ (Google Chrome branded builds) silently blocks `--remote-debugging-port` when the **default** user data directory is used. This breaks every browser-harness workflow that relies on local CDP:
- The "Allow remote debugging?" dialog appears on every Chrome restart
- `DevToolsActivePort` may not be written
- Headless / automation flows fail with cryptic EOF errors

This PR fixes the root cause cross-platform and removes the manual "Allow" step entirely on Linux.

## Issues fixed

- Fixes #107 — Chrome remote debugging was a hidden prerequisite with no guidance
- Fixes #123 — README Quick Start now documents the remote-debugging prerequisite and the Chrome 147 workaround
- Fixes part of #124 — the Chrome 147 user-data-dir block on fresh installs (Windows + macOS profile-copy path, Linux CHROME_CONFIG_HOME path)
- Related to #106 — Chrome 147 CDP failures (while #106 is specifically about the `Sec-WebSocket-Extensions` header, the underlying trigger is often the remote-debugging block)

## How it works

Chrome 147+ checks `IsUsingDefaultDataDirectory()` in `browser_process_impl.cc`. If the profile path matches the platform default, remote debugging is blocked unless the user clicks Allow.

**Linux**: `CHROME_CONFIG_HOME` is honoured by Chrome. We set it to a persistent fake path (`~/.local/share/browser-harness/chrome-config-home`). Chrome computes the "default" dir as `$CHROME_CONFIG_HOME/google-chrome`, while `--user-data-dir` still points to the real `~/.config/google-chrome`. The paths differ → check passes → no dialog. Zero data copying.

**macOS / Windows**: `CHROME_CONFIG_HOME` is ignored. We create a one-time *copy* of the real profile to a persistent temp path (`~/Library/Application Support/browser-harness/chrome-profile-copy` on macOS, `%LOCALAPPDATA%\browser-harness\chrome-profile-copy` on Windows) and point `--user-data-dir` there. The copy is refreshed only when the real profile has newer files. Lock files (`SingletonLock`, `DevToolsActivePort`, etc.) are skipped so Chrome doesn't think another instance is running.

## Changes

- **`launch.py`** (new): `launch_chrome()` — cross-platform Chrome launcher
  - Linux: `CHROME_CONFIG_HOME` bypass (no copy)
  - macOS / Windows: temp profile copy with mtime-based freshness check
  - Auto-detects Wayland on Linux (`--ozone-platform=wayland`)
  - Probes `127.0.0.1:9222` + `/json/version` when `DevToolsActivePort` is missing
- **`daemon.py`**: fallback to port probe when `DevToolsActivePort` file is absent (Chrome 147+ with `CHROME_CONFIG_HOME` sometimes skips writing it)
- **`run.py`**: auto-launch Chrome via `launch_chrome()` when no CDP browser is found; clear error messages telling users to close Chrome first if it's already running without remote debugging
- **`admin.py`**: import `launch_chrome` for convenience; `_log_tail()` returns last 5 lines for diagnostics
- **`install.md`**: Chrome 147+ workaround documentation + usage examples
- **`pyproject.toml`**: adds `launch` module

## Testing

- Ubuntu 24.04 + Chrome 147.0.7727.55 (Wayland) — verified no "Allow remote debugging" dialog after first acceptance; `page_info()` and `goto()` end-to-end
- Code paths for macOS / Windows profile-copy logic validated via import and signature checks (full runtime testing on those platforms is pending community verification)

## Backwards compatibility

Existing users who already pass `BU_CDP_WS` or have Chrome running with `--remote-debugging-port` are unaffected. `launch_chrome()` is opt-in via `run.py` auto-launch or explicit `from admin import launch_chrome`.

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -6,6 +6,9 @@ import urllib.request
 from pathlib import Path
 
 
+from launch import launch_chrome  # noqa: F401 — re-exported for convenience
+
+
 def _load_env():
     p = Path(__file__).parent / ".env"
     if not p.exists():
@@ -35,7 +38,8 @@ def _paths(name):
 def _log_tail(name):
     p = f"/tmp/bu-{name or NAME}.log"
     try:
-        return Path(p).read_text().strip().splitlines()[-1]
+        lines = Path(p).read_text().strip().splitlines()
+        return "\n".join(lines[-5:])  # last 5 lines for context
     except (FileNotFoundError, IndexError):
         return None
 
--- a/admin.py
+++ b/admin.py
@@ -7,6 +7,7 @@ from pathlib import Path
 
 
 from launch import launch_chrome  # noqa: F401 — re-exported for convenience
+from transport import cleanup_endpoint, connect_client, runtime_paths, version_cache_path
 
 
 def _load_env():
@@ -26,43 +27,35 @@ _load_env()
 NAME = os.environ.get("BU_NAME", "default")
 BU_API = "https://api.browser-use.com/api/v3"
 GH_RELEASES = "https://api.github.com/repos/browser-use/browser-harness/releases/latest"
-VERSION_CACHE = Path("/tmp/bu-version-cache.json")
+VERSION_CACHE = version_cache_path()
 VERSION_CACHE_TTL = 24 * 3600
 
 
 def _paths(name):
     n = name or NAME
-    return f"/tmp/bu-{n}.sock", f"/tmp/bu-{n}.pid"
-
-
+    paths = runtime_paths(n)
+    return str(paths.sock), str(paths.pid)
 def _log_tail(name):
-    p = f"/tmp/bu-{name or NAME}.log"
+    p = runtime_paths(name or NAME).log
     try:
-        lines = Path(p).read_text().strip().splitlines()
+        lines = p.read_text().strip().splitlines()
         return "\n".join(lines[-5:])  # last 5 lines for context
     except (FileNotFoundError, IndexError):
         return None
-
-
 def daemon_alive(name=None):
     try:
-        s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-        s.settimeout(1)
-        s.connect(_paths(name)[0])
+        s = connect_client(name or NAME, timeout=1)
         s.close()
         return True
-    except (FileNotFoundError, ConnectionRefusedError, socket.timeout):
+    except (FileNotFoundError, OSError, ValueError):
         return False
-
-
 def ensure_daemon(wait=60.0, name=None, env=None):
     """Idempotent. Self-heals stale daemon, cold Chrome, and missing Allow on chrome://inspect."""
     if daemon_alive(name):
         # Stale daemons accept connects AND reply to meta:* (pure Python) even when the
         # CDP WS to Chrome is dead — probe with a real CDP call and require "result".
         try:
-            s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM); s.settimeout(3)
-            s.connect(_paths(name)[0])
+            s = connect_client(name or NAME, timeout=3)
             s.sendall(b'{"method":"Target.getTargets","params":{}}\n')
             data = b""
             while not data.endswith(b"\n"):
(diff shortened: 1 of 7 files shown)


1RB:
All cubic review issues addressed:

1. **Wayland detection** - now checks `WAYLAND_DISPLAY` / `XDG_SESSION_TYPE` env vars first, then queries only the current logind session (via `XDG_SESSION_ID` or `loginctl show-user`), removed the `pgrep gnome-shell` scan that could falsely match other sessions.

2. **Chrome stderr pipe** - redirected to a temp log file instead of `subprocess.PIPE`, so the pipe buffer can never block the browser process.

3. **CDP exception safety** - `_probe_cdp_port()` now wraps the `/json/version` fetch + JSON parse in a broad `except Exception` that re-raises as `RuntimeError`, so `get_ws_url()` always sees the expected exception type.

4. **pgrep FileNotFoundError** - `_chrome_is_running()` now catches `FileNotFoundError` alongside `CalledProcessError`, so the recovery path doesn't crash on systems without `pgrep` installed.

Qodo-Free-For-OSS:
Hi, `launch_chrome()` computes freshness by `rglob()`+`stat()` over the entire profile tree without handling `OSError`, so permission issues or concurrently-changing files can crash Chrome auto-launch on macOS/Windows.

**How to fix:** Guard rglob/stat with try

> ### Issue description
> `launch.launch_chrome()` can crash on macOS/Windows while computing profile freshness because it does a full recursive `rglob()` and `stat()` without handling filesystem exceptions.
> 
> ### Issue Context
> This runs before launching Chrome (macOS/Windows profile-copy strategy) and will fail on common conditions (locked files, permission-restricted entries, files disappearing during traversal).
> 
> ### Fix Focus Areas
> - launch.py[213-238]
> 
> ### Suggested approach
> - Replace the `max(p.stat().st_mtime for p in ...rglob(...))` with a safe walker that catches `OSError` per-entry.
> - Consider using a cheaper freshness signal (e.g., compare a small set of known files, or store a previous snapshot timestamp) to avoid O(N) scans on every run.

---

1RB:
@qodo-ai-reviewer Thanks for the detailed review! I've fixed the launch.py crash by adding a safe walker that catches OSError per entry. Please share any other issues you've noticed.

Alezander9:
Closing, I believe this issue was fixed in https://github.com/browser-use/browser-harness/pull/265

PR #170, opened 2026-04-23
docs: lead README with first-buyer job

### LINKED ISSUE #90: README framing suggestion: lead with the first operator job
This looks real fast.

The only place I’d push harder is the first-buyer story.

Right now the README opens with total freedom, self-healing, and “you will never use the browser again,” which is memorable, but it makes me infer the actual use case.

The sharper wedge seems simpler:
connect an LLM to your real browser with a thin CDP harness when the normal browser-agent stack is too heavy or too opinionated.

I’d try an opener closer to:
“Connect an LLM directly to your real browser with a thin, editable CDP harness — for browser tasks where you need freedom, not another framework.”

Then the self-healing angle becomes proof instead of doing all the positioning work.

If useful, I can sketch a tighter top section around that first operator use case.

mvanhorn:
Per #90, reframe the README opener to name the job (thin, editable CDP harness, freedom over framework) before claiming outcomes. Self-healing stays as a one-liner under the lede so the ASCII demo that follows reads as proof.

@sauravpanda invited this on the issue.

Fixes #90

---
## Summary by cubic
Rewrote the README intro to lead with the job: a thin, editable CDP harness that connects an LLM to a real browser and favors freedom over frameworks. Self-healing is now a one‑liner under the lede so the ASCII demo reads as proof.

<sup>Written for commit 5d4ae44fdc833b569552494393c460c4f6a05ac1. Summary will update on new commits.</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 # Browser Harness ♞
 
-The simplest, thinnest, **self-healing** harness that gives LLM **complete freedom** to complete any browser task. Built directly on CDP.
+Connect an LLM directly to your real browser with a thin, editable CDP harness. For browser tasks where you need freedom, not another framework.
 
-The agent writes what's missing, mid-task. No framework, no recipes, no rails. One websocket to Chrome, nothing between.
+One websocket to Chrome, nothing between. The agent writes what's missing, mid-task; the harness heals itself when it doesn't.
 
 ```
   ● agent: wants to upload a file


Alezander9:
This PR got stale, but I like this rewrite. Am merging a version now. Got any more ideas or feedback @mvanhorn ?

mvanhorn:
Glad this landed - thanks for picking up the rebase. I'll keep an eye on the README and ping if I spot more first-buyer-flow gaps worth filling in.

PR #180, opened 2026-04-24
feat(transport): cross-platform IPC with TCP loopback on Windows

MICORT:
## Summary

Native Windows support for the daemon. Today `asyncio.start_unix_server` raises `NotImplementedError` on Windows and every path is hardcoded under `/tmp/`, so the daemon crashes before listening and `browser-harness --doctor` traceback's out. This PR:

- adds `transport.py` — a small module that owns OS-specific state paths and IPC server/client creation. AF_UNIX on POSIX, AF_INET on `127.0.0.1:<kernel-assigned-port>` on Windows (loopback-only bind — no new deps, no named pipes, no pywin32).
- routes `daemon.py`, `admin.py`, `helpers.py` through it. **Zero `sys.platform` checks outside `transport.py`.**
- writes an atomic endpoint descriptor (`%TEMP%\bu-<NAME>.endpoint` / `/tmp/bu-<NAME>.endpoint`):
  ```json
  {""v"": 1, ""kind"": ""tcp"", ""host"": ""127.0.0.1"", ""port"": 52137}
  ```
  `v: 1` keeps the shape forward-compatible (named pipes, per-name tokens, etc. can opt-in without breaking clients). Written after the listener binds, cleaned up in `finally` and on stale-daemon detection.
- migrates `log` / `pid` / `version-cache` paths to `transport.state_dir()` — they were hardcoded `/tmp/bu-*`, which meant the daemon crashed on Windows at `open(LOG, 'w')` before it even tried to listen.
- detaches `subprocess.Popen` correctly per OS: `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` on Windows, `start_new_session=True` on POSIX.
- guards `os.chmod(SOCK, 0o600)` (no-op on TCP — there's no file, and loopback-only bind IS the access control).
- `--doctor` gains a `transport` row:
  ```
  transport         tcp 127.0.0.1:52137 (name=default)
  ```
  or, on POSIX with the daemon running:
  ```
  transport         uds /tmp/bu-default.sock (name=default)
  ```
- `helpers.screenshot()` default path moves from `/tmp/shot.png` to `transport.state_dir() / 'shot.png'` so it doesn't crash on Windows.

`run.py` is not touched (per the design constraint in `SKILL.md`).

### Security

[...]

PR #192, opened 2026-04-24
feat(transport): cross-platform IPC — rebased #180 on current main

midirectiekade14-debug:
## Summary

Rebases @MICORT's PR #180 (`feat(transport): cross-platform IPC with TCP loopback on Windows`) onto current `main`, with one merge conflict resolved, and verifies all transport tests pass on Windows 11 + Python 3.14.

**Credit belongs to @MICORT** — the transport module, tests, and Windows bootstrap logic are entirely their work. This PR exists only because #180 went `mergeable: false` after `main` moved on, and it's useful to have a ready-to-merge version. Maintainers: happy to close this in favor of #180 being refreshed by the original author.

## What changed vs. #180

One conflict in `helpers.py`:

- `main` renamed `screenshot()` → `capture_screenshot()` (breaking API — used by 20+ domain-skills files)
- #180's version introduced a platform-aware default path via `transport.state_dir()`

Resolution keeps the `capture_screenshot` name (`main`'s API) **and** the platform-aware default (`#180`'s fix):

```python
def capture_screenshot(path=None, full=False):
    if path is None:
        path = str(transport.state_dir() / "shot.png")
    ...
```

`SKILL.md` conflict was resolved by taking `main`'s version (the doc reorganization that happened after #180 was opened).

## Verification

On Windows 11 Pro, Python 3.14.3:

```
$ uv run python -m pytest tests/
======================== 24 passed, 1 skipped in 1.09s ========================
```

End-to-end smoke test against local Chrome (remote-debugging on port 9222):

```python
new_tab('https://example.com')
wait_for_load()
capture_screenshot(path='C:/tmp/bh.png')
page_info()
# → {'url': 'https://example.com/', 'title': '🟢 Example Domain', 'w': 929, 'h': 917, ...}
```

`browser-harness --doctor` reports `transport tcp 127.0.0.1:<port>` correctly; previously it traceback'd on `socket.AF_UNIX` missing.

## Windows install notes (not code changes — just observations for maintainers)

Two things I had to work around locally but that aren't in-scope for this PR:

1. Python 3.14 on Windows prints 🟢 with cp1252 by default → `print(page_info())` crashes with `UnicodeEncodeError`. Workaround: `PYTHONIOENCODING=utf-8`. Could be fixed in `run.py` with `sys.stdout.reconfigure(encoding='utf-8')`.
2. The daemon subprocess (`uv run daemon.py`) needs `uv` on PATH. On Windows that path (`%APPDATA%\Python\Python3xx\Scripts\`) isn't always on the default user PATH.

Happy to submit these as a follow-up PR if useful.

---
## Summary by cubic
Introduces a cross-platform IPC transport: UDS on macOS/Linux and loopback TCP on Windows. Centralizes paths, server/client creation, and process spawn logic in `transport.py`, fixing Windows startup and improving diagnostics.

- **New Features**
  - New `transport.py` exposes endpoint discovery and IPC: descriptor JSON `{v:1, kind, host|path, port}` with atomic writes and cleanup.
  - Windows binds `127.0.0.1:<ephemeral>`; POSIX uses UDS with `0600` perms.
  - `--doctor` now shows transport and endpoint for the current `BU_NAME`.
  - `capture_screenshot()` uses `transport.state_dir()` when no path is given.

- **Bug Fixes**
  - Daemon/log/pid/version-cache paths use `%TEMP%` on Windows (no `/tmp` dependency).
  - Detached spawn uses Windows `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP`; POSIX keeps `start_new_session=True`.
  - On start/restart, stale endpoint/socket files are removed to avoid bind failures.
  - `admin.ensure_daemon`/`restart_daemon` and `daemon` use `transport` for probing, endpoint writes, and cleanup.

<sup>Written for commit 2accac42cadb23bee8e0adb367b90dd1419cf938. Summary will update on new commits.</sup>

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -154,7 +154,7 @@ The *durable* shape of the site — the map, not the diary. Focus on what the ne
 ## Architecture
 
 ```text
-Chrome / Browser Use cloud -> CDP WS -> daemon.py -> /tmp/bu-<NAME>.sock -> run.py
+Chrome / Browser Use cloud -> CDP WS -> daemon.py -> local IPC (UDS on POSIX, TCP loopback on Windows) -> run.py
 ```
 
 - Protocol is one JSON line each way.
@@ -182,7 +182,7 @@ Chrome / Browser Use cloud -> CDP WS -> daemon.py -> /tmp/bu-<NAME>.sock -> run.
   `restart_daemon()`
   `PY`
   before assuming setup is wrong.
-- If `restart_daemon()` also hangs, kill Chrome entirely (`pkill -9 -f "Google Chrome"`), clean sockets (`rm -f /tmp/bu-default.sock /tmp/bu-default.pid`), reopen Chrome (`open -a "Google Chrome"`), wait 5s, then reconnect. This resets all CDP state.
+- If `restart_daemon()` also hangs, kill Chrome entirely (POSIX: `pkill -9 -f "Google Chrome"`; Windows: `taskkill /F /IM chrome.exe`), clean stale state in `$TMPDIR`/`%TEMP%` (`bu-default.endpoint`, `bu-default.sock` on POSIX, `bu-default.pid`), reopen Chrome, wait 5s, then reconnect. This resets all CDP state.
 - Browser Use API is camelCase on the wire. `cdpUrl`, `proxyCountryCode`, etc.
 - Remote `cdpUrl` is HTTPS, not ws. Resolve the websocket URL via `/json/version`.
 - Stop cloud browsers with `PATCH /browsers/{id}` + `{\"action\":\"stop\"}`.
--- a/admin.py
+++ b/admin.py
@@ -5,6 +5,8 @@ import time
 import urllib.request
 from pathlib import Path
 
+import transport
+
 
 def _load_env():
     p = Path(__file__).parent / ".env"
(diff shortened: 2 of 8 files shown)


Qodo-Free-For-OSS:
Hi, daemon.serve() assumes transport.read_endpoint() returns a dict and calls desc.get()/indexes keys; if the endpoint file is missing or malformed it returns None and the daemon crashes during startup logging. This can turn a recoverable endpoint-file issue into a fatal daemon startup failure.

**How to fix:** Guard/validate endpoint descriptor

> ### Issue description
> `daemon.py` reads the endpoint descriptor back from disk after `start_ipc_server()` and assumes it is a dict with required keys. `transport.read_endpoint()` can return `None` (missing file, malformed JSON), and even when it returns a dict it may be missing keys, causing daemon startup to crash.
> 
> ### Issue Context
> This crash happens in the startup path, right after starting the IPC server. Even if the server is actually running, the daemon can exit due to the logging line.
> 
> ### Fix Focus Areas
> - daemon.py[217-221]
>   - Handle `desc is None` defensively (e.g., log a fallback like "listening (endpoint unreadable)"), or avoid re-reading from disk and instead have `start_ipc_server()` return the descriptor alongside the server.
>   - Avoid direct indexing (`desc['host']`) without checking key presence.
> - transport.py[113-135]
>   - (Optional) return `(server, descriptor)` or provide a helper that returns the descriptor used for writing.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source). Free code review for open-source maintainers.*

Alezander9:

PR #196, opened 2026-04-25
fix: support stdin scripts and preserve JS IIFE returns

Lhy099:
## Summary
  - Allow browser-harness to execute scripts from stdin/heredoc when no -c flag is provided
  - Preserve existing -c behavior without reading stdin
  - Make helpers.js() retry IIFE wrapping only after an Illegal return statement error, so valid IIFEs keep their return value

---
## Summary by cubic
Adds stdin/heredoc script support to `browser-harness` when no `-c` is provided. Tightens `helpers.js()` to only auto-wrap on "Illegal return statement" when it’s a SyntaxError, preserving valid IIFE return values.

- **New Features**
  - `browser-harness` executes code from stdin when run with no args.
  - `-c` behavior is unchanged and never reads stdin.
  - Prints a concise usage message for invalid or empty input.

- **Bug Fixes**
  - `helpers.js()` retries with an IIFE only when "Illegal return statement" is a SyntaxError; runtime Error messages are not retried.
  - Valid IIFEs are evaluated as-is and keep their return value; simple expressions still pass through.

<sup>Written for commit 9b9fb482a39e428d988b3420f1fba6aa52507fda. Summary will update on new commits.</sup>

### DIFF
--- a/helpers.py
+++ b/helpers.py
@@ -202,6 +202,17 @@ def _exception_text(r):
     e = d.get("exception") or {}
     return "\n".join(str(x) for x in (d.get("text"), e.get("description"), e.get("value")) if x)
 
+def _is_illegal_return_syntax_error(r):
+    d = r.get("exceptionDetails") or {}
+    e = d.get("exception") or {}
+    text = str(d.get("text") or "")
+    description = str(e.get("description") or "")
+    class_name = e.get("className")
+    return (
+        "Illegal return statement" in (text + "\n" + description)
+        and (class_name == "SyntaxError" or "SyntaxError:" in text or "SyntaxError:" in description)
+    )
+
 def js(expression, target_id=None):
     """Run JS in the attached tab (default) or inside an iframe target (via iframe_target()).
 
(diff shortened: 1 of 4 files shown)


Qodo-Free-For-OSS:
Hi, helpers.js() retries evaluation whenever the response looks like a SyntaxError with the substring "Illegal return statement", which can also occur when user code throws a SyntaxError at runtime, causing the script to execute twice (duplicating side effects). This retry cannot fix such runtime-thrown errors and only increases risk/cost.

**Severity:** remediation recommended | **Category:** reliability

**How to fix:** Gate retry on return presence

> ## Issue description
> `helpers.js()` retries evaluation (wrapped in an IIFE) based solely on exception text/className, which can be triggered by user code that *throws* a `SyntaxError` at runtime (with the same message). That causes double execution and duplicated side effects, and wrapping cannot resolve the underlying runtime exception.
> 
> ## Issue Context
> The retry is intended only for the *parse-time* "Illegal return statement" from top-level `return`.
> 
> ## Fix Focus Areas
> - helpers.py[205-226]
>   - Add an additional guard before retrying, e.g. require the original `expression` to plausibly contain a top-level return (`"return" in expression` at minimum), or otherwise narrow the predicate so runtime-thrown errors don’t trigger a retry.
>   - (Optional) Add a unit test that simulates a runtime-thrown `SyntaxError` with "Illegal return statement" and asserts no retry occurs.
> - test_js.py[33-84]
>   - Extend coverage for the new guard behavior.

---

PR #203, opened 2026-04-26
fix: scan Chrome profile subdirs for DevToolsActivePort

j17drake:
## Summary
- Scans Chrome profile subdirectories like `Default` and `Profile 1` for `DevToolsActivePort`
- Keeps the existing top-level User Data lookup as a candidate
- Handles missing, non-directory, or permission-denied profile paths without aborting the scan

## Why
Chrome can write `DevToolsActivePort` inside an active profile directory instead of directly under the User Data root. When that happens, browser-harness cannot attach even though Chrome remote debugging is available.

## Verification
- `python3 -m py_compile daemon.py`
- Tested in a fork where Chrome profile subdirectory scanning was needed

---
## Summary by cubic
Fix Chrome attach by scanning profile subdirectories (Default, Profile 1, etc.) for `DevToolsActivePort` instead of only the User Data root. Avoids failures when Chrome writes the port file inside a profile.

- **Bug Fixes**
  - Add candidate search across profile dirs; keep root as a candidate.
  - Treat missing, non-directory, and permission-denied paths as skippable.
  - Probe each candidate port and return the first live `ws://` URL.

<sup>Written for commit 4f15173c777b2617011c7b7f278f3559c3f4cb0f. Summary will update on new commits.</sup>

### DIFF
--- a/daemon.py
+++ b/daemon.py
@@ -62,26 +62,35 @@ def get_ws_url():
     if url := os.environ.get("BU_CDP_WS"):
         return url
     for base in PROFILES:
+        candidates = [base]
         try:
-            port, path = (base / "DevToolsActivePort").read_text().strip().split("\n", 1)
-        except (FileNotFoundError, NotADirectoryError):
-            continue
-        deadline = time.time() + 30
-        while True:
-            probe = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            probe.settimeout(1)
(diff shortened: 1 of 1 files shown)


Alezander9:
Closing, I believe this issue was fixed in https://github.com/browser-use/browser-harness/pull/265. Thank you for raising this issue

PR #209, opened 2026-04-26
domain-skills/loom: library folder enumeration

everlastconsulting:
## What

Adds `domain-skills/loom/folder-enumeration.md` covering enumeration of Loom **library folders** (private workspace folders, URL pattern `loom.com/looms/videos/<slug>-<id>`).

## Why this is in browser-harness and not yt-dlp

yt-dlp already has `LoomFolderIE`, but it covers only **public-shared** folders (`loom.com/share/folder/<id>`). The underlying `/v1/folders/<id>` REST endpoint returns `Forbidden` for library folder IDs even with the user's authenticated cookies — these are structurally different endpoints in Loom's API.

The library variant is the one every Loom user sees in their own workspace sidebar. There is no public read API for it, and most workspace owners don't want to flip private course content to "anyone with link" just to bulk-export it. The only practical programmatic route is the authenticated session in the user's open browser, which is exactly what browser-harness is for.

## What's in the doc

- URL-pattern decision table (when to use yt-dlp vs this skill)
- The `[data-videoid]` selector and the title-extraction noise to strip
- The big learning: Loom's virtual scroller silently caps `scrollTop`. The reliable mechanic is `lastCard.scrollIntoView({block: "end"})`. Empirical numbers from a 78-video test folder included.
- Dead ends that look promising but aren't (`/v1/folders/<id>` for library IDs, GraphQL query reverse-engineering)
- A pipe-to-yt-dlp section showing a 16-parallel `-f http-transcoded` setup that pulled the test folder (78 videos, ~9 GB at 1080p) in ~6 minutes
- Gotchas section covering the duplicate-tab trap and the title-string ARIA wrapping

## Field-tested

Tested on 2026-04-26 against a real workspace library folder: 78 videos enumerated via the scrollIntoView loop, ~9 GB downloaded at 1080p in ~6 minutes (16-parallel yt-dlp with `-f http-transcoded` after enumeration finished).

---
## Summary by cubic
Adds documentation for enumerating Loom private library folders in the browser and piping results to `yt-dlp` for fast downloads. Clarifies when to use this approach vs `yt-dlp`’s existing public-folder support.

- **New Features**
  - Added `domain-skills/loom/folder-enumeration.md`.
  - Explains when to use this skill vs `yt-dlp` (public share folders vs private library folders).
  - Shows how to attach to the user’s open Loom tab and avoid duplicate tabs.
  - Uses `[data-videoid]` for IDs and strips ARIA noise from titles.
  - Details the virtual scroller quirk and uses `scrollIntoView({block: "end"})` instead of `scrollTop`.
  - Calls out dead-end APIs and provides a parallel `yt-dlp` flow with `-f http-transcoded` for faster downloads.

<sup>Written for commit cba31dfec0693976e758807b321ec14cbbdcc83e. Summary will update on new commits.</sup>

### DIFF
--- a/domain-skills/loom/folder-enumeration.md
+++ b/domain-skills/loom/folder-enumeration.md
@@ -1,0 +1,170 @@
+# Loom — Library Folder Enumeration
+
+Field-tested against loom.com on 2026-04-26.
+For private workspace library folders that require an authenticated session.
+
+## TL;DR — When to use this skill vs yt-dlp
+
+Loom has **two** kinds of folder URLs and they need different tools:
+
+| URL pattern | Visibility | Tool |
+|---|---|---|
+| `loom.com/share/folder/<32-hex>` | Public-shared (anyone with link) | **yt-dlp** — `LoomFolderIE` already supports this. Skip browser-harness entirely. |
+| `loom.com/looms/videos/<slug>-<32-hex>` | Private workspace library | **browser-harness** (this skill). yt-dlp doesn't support library folders, and the underlying `/v1/folders/<id>` endpoint returns `Forbidden` even with cookies. |
+
+The library variant is what every Loom user sees in their own workspace sidebar. There is no public read API for it; the only programmatic route is the authenticated session in the user's browser. That's why this skill exists.
+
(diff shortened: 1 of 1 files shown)


sauravpanda:
Thanks alot!

sauravpanda (approved):

PR #225, opened 2026-04-27
Windows support: route IPC through ipc.py (TCP on Windows, AF_UNIX on POSIX)

### LINKED ISSUE #124: Windows support: 4 issues found during fresh install (AF_UNIX, /tmp paths, emoji encoding, Chrome 147 user-data-dir)

Alezander9:
## Summary

The harness was Linux/macOS-only because daemon IPC hardcoded `AF_UNIX` sockets at `/tmp/bu-*.sock` paths and `asyncio.start_unix_server`. On Windows none of these work, and `uv`-managed Python (python-build-standalone) ships without `socket.AF_UNIX` entirely.

This PR centralizes the platform fork in a new `ipc.py`:

- **POSIX**: `AF_UNIX` socket at `<tempdir>/bu-<NAME>.sock` (chmod 0600). Logically equivalent to the prior code, just relocated through one helper.
- **Windows**: TCP loopback on `127.0.0.1:<ephemeral port>`, with the chosen port written to `<tempdir>/bu-<NAME>.port` so clients can find the daemon. Uses stdlib `asyncio.start_server` — no third-party deps, no obscure APIs.

All log/pid/port files now sit under `tempfile.gettempdir()` so they land in `/tmp` on Linux, `$TMPDIR` on macOS, `%TEMP%` on Windows. `helpers.screenshot()` default path moved similarly.

`subprocess` detach uses `start_new_session=True` on POSIX and `DETACHED_PROCESS|CREATE_NEW_PROCESS_GROUP` on Windows via `ipc.spawn_kwargs()`.

`run.py` reconfigures stdout to UTF-8 on Windows so `print(page_info())` doesn't `UnicodeEncodeError` on the 🟢 marker helpers prepend to tab titles.

## Closes / addresses

- Closes #124 item 1 (`AF_UNIX` not available in uv-managed Python)
- Closes #124 item 2 (hardcoded `/tmp` paths)
- Closes #124 item 4 (`UnicodeEncodeError` on cp1252 stdout)
- Item 3 (Chrome 147 user-data-dir) is a separate concern, not addressed here

Thanks to @seltzdesign for the detailed Windows triage in #124. The TCP-on-Windows direction was inspired by their proposal.

## Why TCP on Windows, not named pipes

I prototyped named pipes first (`asyncio.start_serving_pipe` + a custom `_PipeClient`) and verified it worked end-to-end, but it required:

- A custom `asyncio.Protocol` subclass adapting raw pipe transport to the `(reader, writer)` handler shape `asyncio.start_unix_server` provides automatically.
- Inheriting from `asyncio.streams.FlowControlMixin` to make `StreamWriter.drain()` work — that bug only surfaces under named pipes because so few people use the pipe API.
- A separate `_PipeClient` wrapping a binary file handle to emulate a socket interface for blocking clients.

TCP on Windows + AF_UNIX on POSIX is **simpler** (50 LoC of logic vs. 80), uses **only the most boring stdlib asyncio APIs**, and `connect()` returns a real `socket` on both platforms instead of a wrapper. The trade-off is a port allocation per daemon and a `.port` sidecar file; in exchange we drop ~60 lines of platform-specific glue.

## Verification

**Windows 11 Pro + Chrome 147 + uv-managed Python 3.12** — full end-to-end:

```powershell
> uv run python -c "from admin import ensure_daemon; ensure_daemon(); print('up')"
up
> uv run python -c "from helpers import goto, page_info, screenshot; import time; goto('https://example.com'); time.sleep(2); print(page_info()); print(screenshot())"
{'url': 'https://example.com/', 'title': '🟢 Example Domain', 'w': 2560, 'h': 1295, 'sx': 0, 'sy': 0, 'pw': 2560, 'ph': 1295}
C:\Users\alexa\AppData\Local\Temp\shot.png
> Get-Item $env:TEMP\shot.png | Select-Object Length

Length
------
 20352

> uv run python -c "from admin import restart_daemon; restart_daemon(); print('stopped')"
stopped
> Test-Path $env:TEMP\bu-default.pid
False
```

**POSIX path**: not yet tested in this branch, but is logically equivalent to the previous code (same `AF_UNIX` call, same `chmod 0600`, same socket-file semantics) routed through `ipc.py`'s POSIX branch. **One observable behavior change on macOS**: socket/log/pid/screenshot files now land in `$TMPDIR` (`/var/folders/...`) instead of `/tmp`. They all move together; any caller hardcoding `/tmp/bu-*` paths would need updating.

## Files changed

- `ipc.py` (new) — platform fork
- `daemon.py` — replace `/tmp/...` constants and `AF_UNIX` plumbing with `ipc.*` calls
- `admin.py` — same: `_paths`, `_log_tail`, `daemon_alive`, `restart_daemon`. Also swap `start_new_session=True` for `**ipc.spawn_kwargs()`
- `helpers.py` — `_send` socket → `ipc`, `screenshot()` default path → `tempfile.gettempdir()`
- `run.py` — Windows UTF-8 stdout reconfigure
- `pyproject.toml` — add `ipc` to `py-modules`
- `SKILL.md`, `interaction-skills/connection.md` — update `/tmp/bu-*` references to be cross-platform


Alezander9:
Thanks @cubic-dev-ai. Addressed all four findings in b749f82:

- **daemon.py:1, SKILL.md:151** — docstring/diagram updated to say TCP loopback. Took the suggested wording.
- **daemon.py:209** — fixed the missed-serve-crash issue, but with a tighter shape than the suggestion. Race `serve_task` and `stop_task` via `asyncio.wait(FIRST_COMPLETED)`; if `serve_task` finishes first it must have raised (it never returns on success), so `await` it to surface the exception. `finally` cancels both unconditionally. Same correctness guarantee, less machinery, no early `done()` check needed because the wait covers both states.
- **ipc.py:45** — added `_check(name)` validating `^[A-Za-z0-9_-]{1,64}$`, called from every path builder (`log_path`/`pid_path`/`port_path`/`_sock_path`). Single point of validation, bad names raise `ValueError` early.

Re-tested name validation, serve-crash detection, and the normal shutdown path. End-to-end Windows + POSIX still pass.

PR #227, opened 2026-04-28
fix(windows): add cross-platform daemon transport fallback

HzaCode:
## Summary

This PR makes the daemon transport work on Windows while preserving the existing Unix domain socket behavior on macOS/Linux.

### Changes

- Add `transport.py` to centralize daemon socket handling.
- Keep POSIX behavior unchanged:
  - `/tmp/bu-{NAME}.sock`
  - Unix domain sockets via `AF_UNIX`
  - socket file permission `0o600`
- Add Windows fallback:
  - TCP loopback on `127.0.0.1`
  - deterministic port derived from `BU_NAME`
  - temp-dir based pid/log paths
- Update `admin.py`, `daemon.py`, and `helpers.py` to use the transport abstraction.
- Fix Windows daemon lifecycle issues:
  - use `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` instead of relying on POSIX-style `start_new_session`
  - avoid `os.kill(pid, 0)` on Windows, where it raises `WinError 87`
- Add transport unit tests to lock in both POSIX and Windows behavior.

## Testing

On Windows 10 / Python 3.11.13:

```text
uv run --with pytest pytest -v --tb=short
# 13 passed
```

Manual smoke tests:

```text
uv run browser-harness --reload
uv run browser-harness --setup
uv run browser-harness --doctor
uv run browser-harness -c print(page_info())
```

Observed:

- `--setup` starts the daemon successfully.
- `--doctor` reports both Chrome and daemon as OK.
- `page_info()` successfully returns the active browser page metadata through the daemon.

## Compatibility notes

This intentionally keeps macOS/Linux on the original Unix socket path. The new tests assert that POSIX-style usage still calls `AF_UNIX` and connects to `/tmp/bu-{name}.sock`.

On Windows, the fallback listens only on `127.0.0.1`. This is a pragmatic replacement for Unix domain sockets, but unlike the POSIX `0o600` socket file, TCP loopback does not provide per-user filesystem permissions. This should be fine for the common single-user desktop workflow, but it is worth noting for multi-user Windows hosts.

---
## Summary by cubic
Adds a cross‑platform daemon transport that keeps Unix sockets on macOS/Linux and falls back to TCP loopback on Windows. Also fixes Windows daemon start/stop behavior.

- **Bug Fixes**
  - Added `transport.py` to centralize endpoints and cleanup; updated `admin.py`, `daemon.py`, and `helpers.py` to use it; added unit tests.
  - POSIX unchanged: `/tmp/bu-{NAME}.sock` via `AF_UNIX` with `0o600` perms.
  - Windows fallback: TCP `127.0.0.1` with a deterministic port from `BU_NAME`, PID/log in temp dir, proper detach flags (`DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP`), and no `os.kill(pid, 0)` probe.

- **Migration**
  - No action needed on macOS/Linux.
  - On Windows the daemon listens on `127.0.0.1:PORT`; per‑user file permissions are not enforced on TCP, which may matter on multi‑user hosts.

<sup>Written for commit bbf41c138487be2c48430025a4f2240efe2e1716. Summary will update on new commits. <a href="https://cubic.dev/pr/browser-use/browser-harness/pull/227?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -1,9 +1,12 @@
 import json
 import os
 import socket
+import sys
 import time
 import urllib.request
 from pathlib import Path
+
+from transport import BASE, cleanup_endpoint, connect_socket, paths
 
 
 def _load_env():
@@ -23,19 +26,19 @@ _load_env()
 NAME = os.environ.get("BU_NAME", "default")
 BU_API = "https://api.browser-use.com/api/v3"
 GH_RELEASES = "https://api.github.com/repos/browser-use/browser-harness/releases/latest"
-VERSION_CACHE = Path("/tmp/bu-version-cache.json")
+VERSION_CACHE = BASE / "bu-version-cache.json"
 VERSION_CACHE_TTL = 24 * 3600
 
 
 def _paths(name):
-    n = name or NAME
-    return f"/tmp/bu-{n}.sock", f"/tmp/bu-{n}.pid"
+    sock, pid, _ = paths(name)
+    return str(sock), str(pid)
 
 
 def _log_tail(name):
-    p = f"/tmp/bu-{name or NAME}.log"
+    _, _, log_path = paths(name)
     try:
-        return Path(p).read_text().strip().splitlines()[-1]
+        return log_path.read_text().strip().splitlines()[-1]
     except (FileNotFoundError, IndexError):
         return None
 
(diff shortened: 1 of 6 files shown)


Alezander9:

HzaCode:
Thanks for the update — that totally makes sense.

People are opening PRs so quickly because the tool is genuinely valuable. Happy to have contributed some testing and an independent implementation path, even if this PR didn’t need to be merged in the end.



> Windows IPC support has landed in #225 (TCP loopback on Windows, AF_UNIX on POSIX). Closing as superseded — Windows now works on main. Thanks for the contribution!

PR #236, opened 2026-04-28
Fix Windows daemon probing and CLI docs

simi2525:
## Root Cause

The latest `src/browser_harness` layout moved the CLI to `browser-harness -c`, but several setup/domain examples still used the removed stdin heredoc invocation. On Windows, the daemon health probe in `ensure_daemon()` still attempted a POSIX socket connection, so a healthy Windows daemon could be treated as stale and restarted on every call.

## Fix Summary

- Probe live daemons through the shared IPC abstraction so Windows TCP loopback daemons are recognized as healthy.
- Give the Windows `tasklist` check more time before declaring Chrome absent.
- Update setup/profile-sync/domain-skill examples to use the current `browser-harness -c` invocation.
- Add unit coverage for the Windows daemon probe and slower `tasklist` path.

## Verification

- `git diff --check` -> passed
- `uv run --with pytest pytest` -> 35 passed
- `browser-harness --doctor` -> core health OK: Chrome running, daemon alive, 1 active browser connection
- `browser-harness -c "print(page_info())"` -> succeeded against the attached Chrome tab

## Notes

`profile-use` and `BROWSER_USE_API_KEY` remain optional and were not configured in the local environment.

---
## Summary by cubic
Fixes Windows daemon health checks by probing over the shared IPC so healthy loopback daemons aren’t restarted. Updates examples to use `browser-harness -c` and clarifies Windows TCP architecture; improves Chrome detection reliability on Windows.

- **Bug Fixes**
  - Probe daemons via `ipc.connect(...)` instead of a POSIX socket so Windows TCP daemons are treated as healthy; close the probe socket cleanly.
  - Increase Windows `tasklist` timeout to 15s to avoid false “Chrome not running” results.
  - Add unit tests for the IPC probe and the slower Windows `tasklist` path.

- **Migration**
  - Use `browser-harness -c '...'` (or `-c "$(cat <<'PY' ... )"`) in scripts; examples and install docs were updated, including a note that Windows uses 127.0.0.1:<ephemeral> with a temp `bu-<NAME>.port` file.

<sup>Written for commit d41b87e4ea65b542803d40bf4edba9d592c846b6. Summary will update on new commits. <a href="https://cubic.dev/pr/browser-use/browser-harness/pull/236?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/agent-workspace/domain-skills/facebook/groups.md
+++ b/agent-workspace/domain-skills/facebook/groups.md
@@ -171,7 +171,7 @@ print(js("""
 ## Full example — mine one group, emit JSON for downstream tools
 
 ```bash
-cd ~/Developer/browser-harness && uv run browser-harness <<'PY'
+cd ~/Developer/browser-harness && uv run browser-harness -c "$(cat <<'PY'
 import json, sys
 from urllib.parse import urlparse, parse_qs, unquote
 
@@ -223,6 +223,7 @@ print(json.dumps({
     "external_urls": all_externals,
 }, ensure_ascii=False))
 PY
+)"
 ```
 
 The JSON on stdout is the handoff payload — parse it in the calling agent and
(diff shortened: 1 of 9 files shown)


simi2525:
Closing this draft because it was superseded by the fresh ready PR #238, based on the current origin/main.

PR #240, opened 2026-04-28
fix(admin): route ensure_daemon CDP probe through ipc.connect (Windows)

Alezander9:
## The bug

`ensure_daemon` does an "is the existing daemon healthy" probe before deciding whether to spawn a new one. That probe at `admin.py:152-153` used a raw AF_UNIX socket:

```python
s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM); s.settimeout(3)
s.connect(_paths(name)[0])
```

On Windows that's broken two ways:

1. `socket.AF_UNIX` does not exist on the Python build uv ships on Windows (python-build-standalone) — instant `AttributeError`.
2. Even if it did, `_paths(name)[0]` returns `ipc.sock_addr(name)` which on Windows is the display string `"127.0.0.1:<port>"` — not a valid socket path.

Either way: exception → swallowed by `except Exception: pass` → falls through to `restart_daemon(name)` → daemon killed and respawned on every warm call. Slow and churns the Chrome connection. Reported by a Windows user post-#225.

## The fix

Use `ipc.connect(name)` — the platform-aware helper already in `_ipc.py` that `daemon_alive` and `restart_daemon` use. It returns AF_UNIX on POSIX, TCP loopback on Windows. Pure correctness fix; POSIX behavior is functionally identical.

Also dropped the now-unused `_paths()` helper — `restart_daemon` only needed the `pid_path` half, inlined.

## Verification

- `uv run pytest tests/unit` — all 16 pass.
- POSIX: `ipc.connect` returns an AF_UNIX socket, identical to the previous code.
- Windows: `ipc.connect` returns a TCP loopback connection to the port the daemon is actually listening on.

4 lines added, 8 removed in `src/browser_harness/admin.py`.

### DIFF
--- a/src/browser_harness/_ipc.py
+++ b/src/browser_harness/_ipc.py
@@ -3,10 +3,11 @@ import asyncio, os, re, socket, subprocess, sys, tempfile
 from pathlib import Path
 
 IS_WINDOWS = sys.platform == "win32"
-# POSIX: /tmp keeps AF_UNIX paths under sun_path limits (104 on macOS, 108 on Linux).
-# tempfile.gettempdir() on macOS returns /var/folders/... (~49 chars) which combined with
-# a 64-char BU_NAME exceeds the limit. Windows uses TCP, so any tempdir is fine.
-_TMP = Path(tempfile.gettempdir()) if IS_WINDOWS else Path("/tmp")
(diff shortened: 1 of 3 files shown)


Alezander9:
Tested end to end on linux, mac, and windows

PR #251, opened 2026-04-29
Fix remote browser cleanup when daemon startup fails

### LINKED ISSUE #250: `start_remote_daemon` leaves cloud browser running when daemon startup fails
## Bug

`start_remote_daemon()` creates a Browser Use cloud browser before starting the local daemon. If startup fails afterward, the created cloud browser is not explicitly stopped.

## Impact

The remote browser session can be left running even though initialization did not complete.

## Repro shape

1. Call `start_remote_daemon()`
2. Let `/browsers` creation succeed
3. Fail during later startup, such as:
   - CDP URL resolution
   - `ensure_daemon()` startup
4. Observe that cleanup depends on daemon shutdown, which never happens in this failure path

## Expected

If startup fails after remote browser creation, the created cloud browser should be stopped before the error is raised.

## Related

Related to #181 and #183

bittoby:
## Summary

This fixes a remote startup cleanup bug in `start_remote_daemon()`.

If a cloud browser is created successfully but daemon startup fails afterward, the remote browser could be left running because cleanup depended on daemon shutdown, which never happens in that failure path.

## What changed

- stop the created cloud browser when `ensure_daemon()` fails after browser creation
- add unit tests for the failure and success paths

## Why

This avoids leaving cloud browser sessions alive when remote initialization does not complete.

## Testing

```bash
uv run --with pytest pytest tests/unit/test_admin.py -q
```

Passed:
- `14 passed in 0.40s`

Closes #250

---
## Summary by cubic
Fixes remote startup cleanup: if a cloud browser is created but the daemon start fails or is interrupted, we stop the browser to avoid orphaned sessions.

- **Bug Fixes**
  - Wrap `ensure_daemon()` in `start_remote_daemon()` with try/except `BaseException` and call `_stop_cloud_browser()` on error (covers `KeyboardInterrupt`/`SystemExit`).
  - Add `_stop_cloud_browser()` that ignores stop errors and unit tests for failure, interruption, and success paths.

<sup>Written for commit 08fd2494cf2750d69ef405811bea5d12f2d88da1. Summary will update on new commits. <a href="https://cubic.dev/pr/browser-use/browser-harness/pull/251?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/browser_harness/admin.py
+++ b/src/browser_harness/admin.py
@@ -253,7 +253,7 @@ def _stop_cloud_browser(browser_id):
         return
     try:
         _browser_use(f"/browsers/{browser_id}", "PATCH", {"action": "stop"})
-    except Exception:
+    except BaseException:
         pass
 
 
@@ -352,7 +352,7 @@ def start_remote_daemon(name="remote", profileName=None, **create_kwargs):
             name=name,
             env={"BU_CDP_WS": _cdp_ws_from_url(browser["cdpUrl"]), "BU_BROWSER_ID": browser["id"]},
         )
-    except Exception:
+    except BaseException:
         _stop_cloud_browser(browser.get("id"))
         raise
     _show_live_url(browser.get("liveUrl"))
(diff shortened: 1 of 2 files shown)


Crystora:
@MagMueller Could you please review my first PR? I'd appreciate any feedbacks.

sauravpanda (approved):

Crystora:
Thanks 👍

PR #265, opened 2026-04-30
fix(daemon): fall back to DevToolsActivePort ws path when /json/version 404s

### LINKED ISSUE #264: daemon: Chrome 147+ default-profile CDP discovery fails (regression from #260)
## Summary

`get_ws_url()` cannot attach to Chrome 147+ when remote debugging is enabled on the **default** user data directory, even though `DevToolsActivePort` is written and the websocket itself works. The daemon polls `/json/version` for 30s and dies with a misleading *"DevTools is not live yet"* error.

## Root cause

Chrome 147+ silently disables the `/json/*` HTTP discovery endpoints when `--remote-debugging-port` is used with the default user data directory (the `IsUsingDefaultDataDirectory()` check in `browser_process_impl.cc`, the same lockdown discussed in PR #142). On a default-profile Chrome 147 install:

```text
GET /json/version → 404
GET /json         → 404
GET /json/list    → 404
```

…but `DevToolsActivePort` is still written, and the websocket at `ws://127.0.0.1:9222/devtools/browser/<UUID>` works fine.

`daemon.py:get_ws_url()` reads `DevToolsActivePort` to learn the port, then resolves the WS URL via `/json/version` (intentional behavior introduced by **#260**). When `/json/version` 404s permanently, the 30s poll never recovers.

## Why this is a regression

PR #260 fixed the *opposite* failure mode: a stale `DevToolsActivePort` left behind after switching `--user-data-dir`, where the file's stored UUID was dead and `/json/version` had the live one. That fix is correct in isolation, but the new code never falls back to the file's UUID even when the file was just written by the live Chrome.

## Repro (macOS, but same on Win/Linux)

1. Chrome 147 running on the default profile with remote debugging enabled (sticky checkbox from `chrome://inspect/#remote-debugging`, or any prior `--remote-debugging-port=9222` launch).
2. `browser-harness -c 'print(page_info())'`
3. Output:
   ```
   RuntimeError: fatal: Chrome's remote-debugging page is open, but DevTools is not live yet on 127.0.0.1:9222 — if Chrome opened a profile picker, choose your normal profile first, then tick the checkbox and click Allow if shown
   ```
4. Verify the websocket is fine:
   ```bash
   $ cat "~/Library/Application Support/Google/Chrome/DevToolsActivePort"
   9222
   /devtools/browser/14c7583e-...
   $ curl -o /dev/null -w "%{http_code}\n" http://127.0.0.1:9222/json/version
   404
   $ BU_CDP_WS='ws://127.0.0.1:9222/devtools/browser/14c7583e-...' browser-harness -c 'print(page_info())'
   {'url': '...', 'title': '...', ...}    # works
   ```

## Environment

- macOS Darwin 25.4.0, Apple Silicon
- Google Chrome 147.0.7727.137 (default profile, no `--user-data-dir` override)
- browser-harness `d1209ab` (current `main`)
- Python 3.13.11

## Workaround (current users)

Set `BU_CDP_WS=ws://127.0.0.1:9222<path>` from `DevToolsActivePort`'s second line. Persists in `repo_root/.env`.

## Existing PRs that *don't* fix this case

- **#142**, **#166** — both work around Chrome 147's default-profile lockdown by *launching* Chrome on a non-default profile (`CHROME_CONFIG_HOME` or temp profile copy). Doesn't help when Chrome is already running on the default profile and the user has live tabs/state.
- **#127** — handles the inverse: `DevToolsActivePort` missing but `/json/version` live.
- **#203** — scans profile subdirs; this case has the file in the right place.
- **#260 (merged)** — the change that introduced the regression.

## Proposed fix

Read both port and ws path from `DevToolsActivePort`. Try `/json/version` first (preserves #260's stale-UUID behavior). If it returns **HTTP 404 specifically** (vs. connection-refused/timeout, which mean Chrome is still starting), fall back to `ws://127.0.0.1:{port}{ws_path}` from the file. The file's UUID is fresh because Chrome rewrites it on every launch, and the only persistent reason `/json/version` 404s on a live port is the default-profile lockdown.

If the WS upgrade then 404s (the true stale-UUID edge case PR #260 targeted), the existing CDP handshake error surfaces unchanged.

PR opening shortly with the patch + verification on the affected machine.

molesza:
## Summary

Fix CDP attach on Chrome 147+ when remote debugging runs on the **default** user data directory.

Closes #264.

## Problem

Chrome 147+ silently disables the `/json/*` HTTP discovery endpoints when `--remote-debugging-port` is used with the default user data directory (the `IsUsingDefaultDataDirectory()` lockdown — the same one PR #142 sidesteps by launching on a non-default profile). Verified locally on macOS Darwin 25.4 + Chrome 147.0.7727.137:

```text
GET /json/version → 404
GET /json         → 404
GET /json/list    → 404
```

`DevToolsActivePort` is still written, and `ws://127.0.0.1:9222/devtools/browser/<UUID>` still accepts CDP. But `get_ws_url()` reads only the *port* from `DevToolsActivePort` and resolves the WS URL via `/json/version` (intentional behavior introduced in #260). On a default-profile Chrome 147, `/json/version` 404s permanently — the 30s poll never recovers and dies with the misleading `DevTools is not live yet on 127.0.0.1:9222` error.

## Repro

1. Chrome 147 running on the default profile with remote debugging on (sticky checkbox from `chrome://inspect/#remote-debugging` after any prior `--remote-debugging-port=9222` launch).
2. Run any harness command:
   ```
   $ browser-harness -c 'print(page_info())'
   RuntimeError: fatal: Chrome's remote-debugging page is open, but DevTools is not live yet on 127.0.0.1:9222 — ...
   ```
3. Confirm the websocket itself is fine via `BU_CDP_WS` constructed from `DevToolsActivePort`.

## Why #260's concern still holds

PR #260 deliberately stopped trusting the path in `DevToolsActivePort` because a *stale* port file (left from a previous Chrome run with a different `--user-data-dir`) had a dead UUID, while `/json/version` had the live one. That's a real bug and this PR doesn't reintroduce it.

## Fix

Read both port and ws path from `DevToolsActivePort`. Try `/json/version` first — keeps #260's stale-UUID handling intact. On **HTTP 404 specifically** (distinct from connection-refused/timeout, which mean Chrome is still starting), fall back to `ws://127.0.0.1:{port}{ws_path}` from the file. The file's UUID is fresh because Chrome rewrites it on every launch, and the only persistent reason `/json/version` 404s on a live port is the default-profile lockdown. If the fallback path is somehow stale anyway, the existing CDP handshake error surfaces cleanly — no worse than today.

## Behavioral matrix

| Scenario | Before | After |
|---|---|---|
| Chrome 147 default profile (this PR) | ❌ 30s poll + misleading error | ✅ attaches via file path |
| Stale `DevToolsActivePort` from prior `--user-data-dir` (#260's case) | ✅ /json/version returns live UUID | ✅ same path, unchanged |
| Chrome still starting (TCP connect-refused) | ✅ 30s retry on /json/version | ✅ same path, unchanged |
| `BU_CDP_WS` / `BU_CDP_URL` set | ✅ unchanged | ✅ unchanged |
| `DevToolsActivePort` missing entirely | ✅ falls through to 9222/9223 probe | ✅ unchanged |

## Verification

End-to-end on the affected setup:

```
$ rm /tmp/bu-default.* 2>/dev/null
$ unset BU_CDP_WS
$ browser-harness -c 'print(page_info())'
{'url': 'https://...', 'title': '...', 'w': 1900, 'h': 803, ...}

$ tail /tmp/bu-default.log
connecting to ws://127.0.0.1:9222/devtools/browser/14c7583e-755b-4b45-be74-793ea45e81be
attached BBADAE259F24F78824248BED6A3B29F5 ...
listening on /tmp/bu-default.sock (name=default, remote=local)
```

---
## Summary by cubic
Fixes CDP attach on Chrome 147+ when remote debugging uses the default profile. When `/json/version` returns 404, we now fall back to the websocket path from `DevToolsActivePort` to avoid the 30s retry loop and attach immediately.

- **Bug Fixes**
  - Read both port and ws path from `DevToolsActivePort`; try `/json/version` first, and on HTTP 404 only, use `ws://127.0.0.1:{port}{ws_path}`.
  - Keep existing behavior for connection refused/timeouts (keep polling) and for env var overrides; no changes in non-404 scenarios.

<sup>Written for commit 5dfa89fa66c1d75ca3ac10896878aff6ad5aa99f. Summary will update on new commits. <a href="https://cubic.dev/pr/browser-use/browser-harness/pull/265?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/browser_harness/daemon.py
+++ b/src/browser_harness/daemon.py
@@ -105,24 +105,17 @@ def get_ws_url():
         # with a different --user-data-dir on the same port, that file is left behind
         # with a stale browser UUID and the WS upgrade returns 404.
         deadline = time.time() + 30
-        saw_http_404 = False
         while time.time() < deadline:
             try:
                 return json.loads(urllib.request.urlopen(f"http://127.0.0.1:{port}/json/version", timeout=1).read())["webSocketDebuggerUrl"]
             except urllib.error.HTTPError as e:
-                if e.code == 404:
-                    # Chrome 147+ disables /json/* HTTP discovery when remote debugging
-                    # runs on the default user-data-dir (IsUsingDefaultDataDirectory check).
-                    # The websocket itself still works; fall back to the path Chrome wrote
-                    # to DevToolsActivePort, which is fresh because Chrome rewrites it on
-                    # every launch.
-                    saw_http_404 = True
-                    break
+                # Chrome 147+ disables /json/* HTTP discovery on the default user-data-dir;
+                # the ws path Chrome wrote to DevToolsActivePort still works.
+                if e.code == 404 and ws_path:
+                    return f"ws://127.0.0.1:{port}{ws_path}"
                 time.sleep(1)
             except (OSError, KeyError, ValueError):
                 time.sleep(1)
-        if saw_http_404 and ws_path:
-            return f"ws://127.0.0.1:{port}{ws_path}"
         raise RuntimeError(
             f"Chrome's remote-debugging page is open, but DevTools is not live yet on 127.0.0.1:{port} — if Chrome opened a profile picker, choose your normal profile first, then tick the checkbox and click Allow if shown"
         )
--- a/src/browser_harness/daemon.py
+++ b/src/browser_harness/daemon.py
@@ -1,5 +1,5 @@
 """CDP WS holder + IPC relay (Unix socket on POSIX, TCP loopback on Windows). One daemon per BU_NAME."""
-import asyncio, json, os, socket, sys, time, urllib.request
+import asyncio, json, os, socket, sys, time, urllib.error, urllib.request
 from collections import deque
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


Alezander9:
Pushed a small follow-up (fb2571a) that inlines the 404 fallback into the `HTTPError` handler and drops the `saw_http_404` flag.

Same behavioral matrix:
- `/json/version` returns 200 -> early return on first iteration. (#260's stale-UUID protection unchanged.)
- `HTTPError` with `code == 404` and non-empty `ws_path` -> return the file's ws URL immediately. (Chrome 147 default-profile case.)
- `HTTPError` with `code == 404` and empty `ws_path` -> `time.sleep(1)` and keep polling until deadline. (Essentially dead path; Chrome always writes both lines when remote debugging is on.)
- `ConnectionRefusedError` / other `OSError` -> caught by the second except, `time.sleep(1)`, keep polling.
- Non-404 `HTTPError` (e.g. 500) -> falls through to the inner `time.sleep(1)`, keep polling.

Net: -7 lines, no flag, no post-loop branch, comment trimmed to 2 lines. `pytest tests/unit/` -> 25 passed.

Alezander9:
Good work @molesza  , +vouch

Issue #183, opened 2026-04-24
Bug Report: Systemic failure in remote provisioning and daemon lifecycle (5 reproducible failure points) (WHERE IS MY MAC-MINI🧐😂😂)


yousefbassem95-lang:
# ♞ The "Shadow Sentry" Grand Report: 5-Point Structural Failure Analysis

**From:** Youssef
**Date:** April 24, 2026
**Target:** [browser-use/browser-harness](https://github.com/browser-use/browser-harness)

---

## 1. Executive Summary
This report details a systematic breakdown of the `browser-harness` when subjected to high-stress, complex, and non-standard environments. I identified 5 distinct failure points ranging from core daemon initialization bugs to high-level task execution stalls. While I engineered a manual bypass for the connection phase, the tool exhibits critical limitations in handling encapsulated DOM structures and modern Pythonic asynchronous patterns.

---

### Failure 1: The `start_remote_daemon()` Paradox (Daemon Setup Bug)
*   **The Problem:** Calling `start_remote_daemon()` with a valid API key successfully provisions a cloud browser but fails to connect. The tool erroneously falls back to local detection logic.
*   **The Error:** `RuntimeError: fatal: DevToolsActivePort not found in [...]`
*   **The Analysis:** The function fails to correctly prioritize the `BU_CDP_WS` environment variable it just generated, defaulting to searching for local Chrome profile paths that don't exist in headless or containerized environments.
*   **The Fix:** Manually provisioning via `_browser_use`, resolving the `wss://` URL via `/json/version`, and injecting `BU_CDP_WS` directly into the shell before execution.

### Failure 2: The Shadow DOM "Black Hole" (Interaction Bug)
*   **The Problem:** When targeting elements inside a Shadow Root (tested on `whatpumpkin.github.io/ShadowDOM-example/`), the `js()` helper consistently returns `None`.
*   **The Analysis:** Even with direct JavaScript execution, the harness struggles to bridge the gap into encapsulated DOMs. The `None` return value indicates a failure to either locate the element or serialize the returned data from the isolated scope.
*   **Impact:** Renders the harness blind to modern web components and modular UI frameworks.

### Failure 3: The Async Await "Type Collision" (Scripting Bug)
*   **The Problem:** Attempting to run standard `asyncio` code within the harness script block fails immediately.
*   **The Error:** `TypeError: object str can't be used in 'await' expression`
[...]

Issue #181, opened 2026-04-24
Bug: start_remote_daemon fails, falls back to local detection errors, manual BU_CDP_WS workaround


yousefbassem95-lang:
---

**Subject: Bug Report & Workaround: `start_remote_daemon` Failure in Browser Harness**

**From: Youssef**

**Date:** April 24, 2026

---

**1. Executive Summary**

This report details a critical issue encountered with the `browser-harness` tool's automated remote browser provisioning via the `start_remote_daemon` function. Despite correct API credentials and parameters, this function consistently fails to initiate a remote connection and incorrectly returns local browser detection errors. We have identified the root cause as a potential flaw in the `start_remote_daemon` or `ensure_daemon` logic's handling of remote connection parameters, leading to a fallback to local detection. A manual workaround has been developed by directly provisioning a remote browser and setting environment variables, which successfully establishes a connection. This issue represents a significant limitation in the tool's automated setup for users facing environmental complexities.

---

**2. Environment Details**

*   **Tool:** `browser-harness`
*   **Tool Version:** 0.1.0 (git)
*   **Operating System:** Linux (Ubuntu-based)
*   **Python Version:** 3.11.15
*   **Browser Attempts:**
    *   Local Chromium (Snap version, `2:1snap1-0ubuntu2`) - Failed detection.
    *   Playwright's managed Chromium - Failed to locate executable reliably.
*   **Remote Browser Service:** Browser Use API

---

**3. Problem Description**

When attempting to provision a remote browser using the `start_remote_daemon()` function in `admin.py`, the `browser-harness` consistently fails. The process halts with the following error message, which is indicative of a local browser detection failure, even though the intent was to establish a remote connection:

```
fatal: DevToolsActivePort not found in [...] — enable chrome://inspect/#remote-debugging, or set BU_CDP_WS for a remote browser
```

Furthermore, running `browser-harness --doctor` prior to or after these failures consistently shows `[FAIL] daemon alive`, indicating the core daemon process is not starting correctly under these conditions.

---

**4. Steps to Reproduce the Failure**

1.  Clone the `browser-harness` repository.
2.  Install the tool using `uv tool install -e .`.
3.  Obtain a Browser Use API key.
4.  Set the `BROWSER_USE_API_KEY` environment variable in the shell.
5.  Execute the command to start a remote daemon:
    ```bash
    uv run browser-harness <<'PY'
    from admin import start_remote_daemon
    start_remote_daemon("work", proxyCountryCode="us")
[...]

PR #266, opened 2026-04-30
fix: auto-bootstrap cloud daemon on headless servers when BROWSER_USE_API_KEY is set

shaunandrewjackson1977:
## Summary

On headless servers (VPS, Docker) with no local Chrome, `browser-harness -c '...'` always fails even when `BROWSER_USE_API_KEY` is set and Browser Use Cloud is available.

## Repro

1. Run `browser-harness` on a headless Linux server with no Chrome installed
2. Set `BROWSER_USE_API_KEY` to a valid Browser Use Cloud key
3. Run any `-c` script — e.g. `browser-harness -c 'print(page_info())'`

```
RuntimeError: fatal: DevToolsActivePort not found in [...] — enable
chrome://inspect/#remote-debugging, or set BU_CDP_WS for a remote browser
```

## Why

`run.py` calls `ensure_daemon()` unconditionally before executing any user script. On a headless server `ensure_daemon()` finds no local Chrome and raises immediately — the user's script never runs, so `start_remote_daemon()` can never be called from within a `-c` script.

Issues #181 and #183 report this symptom. The root cause analysis in #181 suggests `start_remote_daemon()` fails to pass `BU_CDP_WS` to `ensure_daemon()` — this is not accurate. `start_remote_daemon()` passes it correctly; it simply never gets the chance to run.

## Fix

Add a pre-check before `ensure_daemon()`: if no daemon is alive, Chrome is not listening on known debugging ports, and `BROWSER_USE_API_KEY` is set, auto-provision a Browser Use cloud browser first.

```python
if not daemon_alive() and not _local_chrome_listening() and os.environ.get("BROWSER_USE_API_KEY"):
    start_remote_daemon(NAME)
ensure_daemon()
```

`_local_chrome_listening()` probes ports 9222 and 9223 with a 0.3s timeout — the same ports already probed in `daemon.py`. This is more precise than `_is_local_chrome_mode()`, which only checks for absence of `BU_CDP_WS` and would incorrectly trigger cloud bootstrap on a local machine where Chrome is running but `BROWSER_USE_API_KEY` is also set (e.g. for profile sync).

## Test

Tested on a headless Hostinger VPS running hermes-agent in Docker (no Chrome installed, `BROWSER_USE_API_KEY` set as a container environment variable).

**Before** (without the fix, same environment):

```bash
$ browser-harness -c "new_tab('https://www.google.com'); wait_for_load(); print(page_info())"
RuntimeError: fatal: DevToolsActivePort not found in /tmp/playwright/chrome-profile-...
— enable chrome://inspect/#remote-debugging, or set BU_CDP_WS for a remote browser
```

**After**:

```bash
$ browser-harness -c "new_tab('https://www.google.com'); wait_for_load(); print(page_info())"
https://live.browser-use.com?wss=https://...
(no local GUI — share the liveUrl with the user)
{'url': 'https://www.google.com/', 'title': 'Google', 'w': 1512, 'h': 770, ...}
```

Cloud daemon auto-provisioned transparently — no manual setup required. `_local_chrome_listening()` correctly returns `False` on this host (ports 9222/9223 are not open), so the bootstrap fires. On a local machine with Chrome running, those ports are open and the bootstrap is correctly skipped.

Unit tests:

```
uv run --with pytest pytest tests/unit/test_run.py -q
6 passed in 0.03s
```

### DIFF
--- a/src/browser_harness/run.py
+++ b/src/browser_harness/run.py
@@ -1,4 +1,4 @@
-import os, sys
+import os, socket, sys
 
 # Windows default stdout encoding is cp1252, which can't encode the 🟢 marker
 # helpers prepend to tab titles (or anything else outside Latin-1). Force UTF-8
@@ -9,6 +9,8 @@ if hasattr(sys.stdout, "reconfigure"):
 
 from .admin import (
     _version,
+    NAME,
+    daemon_alive,
     ensure_daemon,
     list_cloud_profiles,
     list_local_profiles,
@@ -44,6 +46,21 @@ Commands:
 """
 
 
+def _local_chrome_listening():
+    """True if Chrome appears to be running with remote debugging on a known port.
+
+    9222 is Chrome's default CDP remote debugging port; 9223 is the common
+    fallback. Consistent with the same probe in daemon.py.
+    """
+    for port in (9222, 9223):
+        try:
+            socket.create_connection(("127.0.0.1", port), timeout=0.3).close()
+            return True
+        except OSError:
+            pass
+    return False
+
+
 def main():
     args = sys.argv[1:]
     if args and args[0] in {"-h", "--help"}:
@@ -71,6 +88,8 @@ def main():
     if len(args) < 2:
         sys.exit("Usage: browser-harness -c \"print(page_info())\"")
     print_update_banner()
+    if not daemon_alive() and not _local_chrome_listening() and os.environ.get("BROWSER_USE_API_KEY"):
+        start_remote_daemon(NAME)
     ensure_daemon()
     exec(args[1], globals())
 
--- a/src/browser_harness/run.py
+++ b/src/browser_harness/run.py
@@ -1,4 +1,4 @@
-import os, socket, sys
+import os, sys, urllib.request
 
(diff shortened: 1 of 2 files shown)


Alezander9:
Pushed a small follow-up (49355ba) addressing cubic's flag and trimming the test set:

- **Probe /json/version instead of bare TCP.** A socket.create_connection to 9222/9223 succeeds against any process bound to those ports, which would skip the cloud bootstrap and then drop the user back into the original DevToolsActivePort not found error inside `ensure_daemon` — exactly the failure mode this PR is meant to fix. `daemon.py`'s fallback already probes `/json/version`; aligning here makes the check Chrome-specific and lets us drop the new `socket` import.
- **Test trim.** The four added tests mocked every collaborator and re-asserted the literal `if`-condition (a 4-row truth table). Kept the positive `test_cloud_bootstrap_on_headless_server`; replaced the three negatives with a focused `test_local_chrome_listening_rejects_non_chrome` that exercises the probe directly — including the false-positive case cubic raised.

Diff is +16/-48. `pytest tests/unit/test_run.py` -> 4 passed.

PR #299, opened 2026-05-05
_ipc: fix AF_UNIX bind/chmod TOCTOU + symlink-pre-plant race (#298)

### LINKED ISSUE #298: Security: TOCTOU window between bind() and chmod(0o600) on daemon AF_UNIX socket
### Before submitting

- [x] I searched existing issues for duplicates.
- [x] I ran `browser-harness --doctor` and read the output.
- [x] I read the troubleshooting section of `install.md`.
- [x] This is a reproducible bug in browser-harness — not a question, feature request, or `cloud.browser-use.com` issue.

### Summary

In `serve()` on POSIX, the daemon's AF_UNIX socket is created and starts accepting connections before its mode is restricted to `0o600`. There is a small TOCTOU window during which the socket's permission bits are whatever the process `umask` produces. On a default umask (`0o022`) the resulting mode is `0o755`, which Linux does not allow other users to `connect()` against — so the practical attack surface is narrow — but the chmod is acting as the only barrier rather than as defense-in-depth, and the socket lives in the shared `/tmp` directory.

A second, related race exists at the `os.unlink(path)` immediately preceding `start_unix_server`: a local attacker can win that gap with a symlink and influence what gets `bind()`ed.

### Repro

This is a code-audit finding rather than a black-box repro. The window can be observed with `strace`:

1. Start the daemon on Linux: `browser-harness --daemon` (or however the harness launches `serve()`).
2. In another shell: `strace -f -e trace=bind,chmod,unlink -p <pid>` against the parent that runs `serve()`.
3. Observe the ordered syscalls:
   - `unlink("/tmp/bu-<NAME>.sock")`
   - `bind(..., "/tmp/bu-<NAME>.sock", ...)` — socket file now exists at `umask`-derived mode and is `listen()`ing
   - `chmod("/tmp/bu-<NAME>.sock", 0o600)` — restriction applied
4. Between step 3.b and 3.c, the socket is reachable (mode-permitting). With `umask 0` set in the parent shell before launch, mode is `0o777` and any local user with traversal rights on `/tmp` can `connect()` and issue arbitrary CDP commands during the window.

Code reference: [`src/browser_harness/_ipc.py:151-154`](../../src/browser_harness/_ipc.py#L151-L154)

```python
path = str(_sock_path(name))
if os.path.exists(path): os.unlink(path)
server = await asyncio.start_unix_server(handler, path=path)
os.chmod(path, 0o600)
```

`asyncio.start_unix_server` calls `bind()` + `listen()` and (with the default `start_serving=True`) begins accepting connections, all before `os.chmod` runs.

### Environment

- OS: Linux (claim verified by reading source); also relevant to macOS.
- Chrome version: N/A (issue is in IPC plumbing, not Chrome integration).
- browser-harness --version: `main` @ commit `0f300e8` (HEAD at time of audit).
- browser-harness --doctor output: N/A — static code review.

JPette1783:
## Summary

Closes #298 — TOCTOU + symlink-pre-plant races on the daemon's AF_UNIX socket.

- **Bind perms TOCTOU (primary).** `asyncio.start_unix_server` used to bind with the inherited umask, leaving the socket world-accessible (mode `0o777` if `umask=0`) until the follow-up `os.chmod(0o600)` ran. Wrapping the bind in `os.umask(0o077)` (try/finally) guarantees the socket is created mode `0o700` from the kernel side; the existing `chmod` stays as defense-in-depth.
- **Symlink-pre-plant race (secondary).** `os.path.exists()` followed symlinks, so a pre-planted dangling symlink at the socket path skipped the unlink and let `bind()` follow it kernel-side. The fix moves the socket into a `0o700` private parent directory (`<tmp>/bu-<NAME>.d/sock`) created/verified by `_ensure_private_dir` (lstat-based, refuses symlinks / non-dirs / wrong-uid), and replaces `os.path.exists() + unlink` with an unconditional `os.unlink` (which never follows symlinks).

Net effect: once `_ensure_private_dir` returns, only our uid can place anything inside the parent dir, so the subsequent unlink+bind on the socket is race-free; and even before that, `umask` ensures no on-disk permission window.

## Changes

- [src/browser_harness/_ipc.py](src/browser_harness/_ipc.py)
  - Added `stat` import.
  - New `_sock_dir(name)` → `<tmp>/bu-<NAME>.d/`. `_sock_path` now returns `<sock_dir>/sock` (was `<tmp>/bu-<NAME>.sock`).
  - New `_ensure_private_dir(p)`: `mkdir(0o700)`, else lstat-verify it's a real dir owned by us and tighten loose perms; refuse on symlink / non-dir / foreign uid.
  - `serve()` POSIX branch rewritten in 3 explicit steps: ensure private parent → unconditional `os.unlink` (catch `FileNotFoundError`) → `os.umask(0o077)` around `start_unix_server` → `chmod(0o600)`.

- [tests/unit/test_ipc.py](tests/unit/test_ipc.py) — 6 new regression tests (POSIX-only via `skipif`):
  - `_ensure_private_dir`: creates 0o700, tightens loose perms, refuses symlinks, refuses non-directory.
  - `serve` socket bound under `umask=0` ends up mode `0o600` with no looser intermediate state.
  - `serve` unlinks a stale dangling symlink at the socket path (the case `os.path.exists` used to silently pass).

## Compatibility notes

- The on-disk path of the socket changes from `/tmp/bu-<NAME>.sock` to `/tmp/bu-<NAME>.d/sock`. All readers of the socket path go through `_ipc._sock_path()` (verified: `connect`, `cleanup_endpoint`, `sock_addr` — daemon, admin, helpers all route through these), so the change is internal. Stale `.sock` files left over from pre-upgrade daemons are harmless leftover bytes.
- AF_UNIX `sun_path` length budget is unaffected: the new layout adds 2 chars (e.g. `/tmp/bu-default.d/sock` = 22 chars vs. 20), well under the 104/108 macOS/Linux limit. The `BH_TMP_DIR`-isolated case (bare `bu` stem) becomes `<BH_TMP_DIR>/bu.d/sock`, still safe.
- Windows TCP-loopback path is untouched; token-based auth there is unchanged.

## Test plan

- [x] `uv run --with pytest pytest tests/unit/test_ipc.py -v` — 16/16 passing (10 existing + 6 new).
- [x] `uv run --with pytest pytest tests/unit/ -v` — 72/72 passing (no regressions in admin/daemon/helpers/run suites that share the `_ipc` paths).
- [ ] Manual smoke: start daemon under `umask 0`, confirm `/tmp/bu-<NAME>.d` is `drwx------` and `/tmp/bu-<NAME>.d/sock` is `srw-------` from the moment it appears.
- [ ] Manual smoke: pre-plant a dangling symlink at `/tmp/bu-<NAME>.d/sock` after the dir exists; confirm the daemon starts cleanly, the symlink is removed, and the bound path is a real socket.
- [ ] macOS smoke (sun_path budget + `/var/folders` default): confirm no `OSError: AF_UNIX path too long`.
- [ ] Optional: `strace -e trace=bind,chmod,umask` to confirm the bind syscall happens with the tightened umask in effect (no observable wider-mode window).

---
## Summary by cubic
Secure AF_UNIX daemon socket binding by creating it in a private 0o700 dir and tightening permissions during bind to remove TOCTOU and symlink-pre-plant races. Also fixes daemon discovery to match the new socket layout; Windows TCP loopback is unchanged.

- **Bug Fixes**
  - Wrap `asyncio.start_unix_server` with `umask 0o077`, then `chmod 0o600`, closing the bind→chmod TOCTOU.
  - Use a `0o700` private parent dir (verified via `lstat`); refuse symlinks/non-dirs/foreign uid; always `unlink` before bind to defeat pre-planted symlinks.
  - Fix `admin._daemon_endpoint_names()` to discover POSIX endpoints at the new layout (require inner `sock`; ignore empty `.d/`); Windows `.port` discovery unchanged.
  - `_ipc.cleanup_endpoint()` now also `rmdir`s the `.d/` wrapper (best-effort). Added POSIX-only regression tests, a macOS-safe short-`/tmp` fixture, and a discovery test tied to `_ipc._sock_path()`.

- **Migration**
  - Socket path is now `<tmp>/bu-<NAME>.d/sock`. All call sites use `_ipc._sock_path()`, so no action needed; old `.sock` files are harmless. Windows unchanged.

<sup>Written for commit aba1efccf10deb4bb1f4343b01e85be4e25cf156. Summary will update on new commits.</sup>

### DIFF
--- a/src/browser_harness/_ipc.py
+++ b/src/browser_harness/_ipc.py
@@ -1,5 +1,5 @@
 """Daemon IPC plumbing. AF_UNIX socket on POSIX, TCP loopback on Windows."""
-import asyncio, json, os, re, secrets, socket, subprocess, sys, tempfile
+import asyncio, json, os, re, secrets, socket, stat, subprocess, sys, tempfile
 from pathlib import Path
 
 IS_WINDOWS = sys.platform == "win32"
@@ -33,7 +33,11 @@ def _stem(name):  # "bu" when BH_TMP_DIR isolates us, else "bu-<NAME>"
 def log_path(name):   return _TMP / f"{_stem(name)}.log"
 def pid_path(name):   return _TMP / f"{_stem(name)}.pid"
 def port_path(name):  return _TMP / f"{_stem(name)}.port"  # Windows-only: holds {"port","token"} JSON
-def _sock_path(name): return _TMP / f"{_stem(name)}.sock"
+# Socket lives inside a 0o700 private parent dir (created by serve()), not
+# directly in shared /tmp — closes the symlink-pre-plant race on bind() that
+# os.path.exists() would silently traverse.
+def _sock_dir(name):  return _TMP / f"{_stem(name)}.d"
+def _sock_path(name): return _sock_dir(name) / "sock"
 
 
 def _read_port_file(name):
@@ -144,13 +148,50 @@ def identify(name, timeout=1.0):
         except OSError: pass
 
 
+def _ensure_private_dir(p):
+    """mkdir 0o700; if it already exists, verify it's a real directory we own
+    with no group/other perms. Refusing here is the symlink-pre-plant defense:
+    once this returns, only our uid can place files inside, so the subsequent
+    unlink+bind on the socket inside is race-free."""
+    try:
+        os.mkdir(p, mode=0o700)
+        return
+    except FileExistsError:
+        pass
+    # lstat (not stat) so a hostile symlink masquerading as the dir is caught
+    # rather than silently traversed.
+    st = os.lstat(p)
+    if not stat.S_ISDIR(st.st_mode):
+        raise RuntimeError(f"{p}: not a directory (refusing — possible symlink attack)")
+    if st.st_uid != os.geteuid():
+        raise RuntimeError(f"{p}: owned by uid {st.st_uid}, expected {os.geteuid()} (refusing)")
+    if st.st_mode & 0o077:
+        os.chmod(p, 0o700)
+
+
 async def serve(name, handler):
     """Run the server until cancelled. handler(reader, writer) sees the same interface either way."""
     global _server_token
     if not IS_WINDOWS:
+        # Step 1: lock down the parent dir (0o700, owned by us). Until this
+        # succeeds, anything inside _TMP is in shared /tmp where attackers
+        # could plant symlinks at our socket path.
+        _ensure_private_dir(_sock_dir(name))
         path = str(_sock_path(name))
-        if os.path.exists(path): os.unlink(path)
-        server = await asyncio.start_unix_server(handler, path=path)
+        # Step 2: unconditional unlink (catches stale symlinks too — os.unlink
+        # never follows them, unlike os.path.exists()).
+        try: os.unlink(path)
+        except FileNotFoundError: pass
+        # Step 3: restrictive umask during bind() closes the TOCTOU window
+        # where the socket would otherwise exist on disk with the inherited
+        # umask (0o777 if umask=0) until chmod() ran. With umask 0o077 the
+        # socket is created mode 0o700 from the start; chmod below tightens
+        # to 0o600 and acts as defense-in-depth.
+        old_umask = os.umask(0o077)
+        try:
+            server = await asyncio.start_unix_server(handler, path=path)
+        finally:
+            os.umask(old_umask)
         os.chmod(path, 0o600)
         _server_token = None
         async with server: await asyncio.Event().wait()
--- a/src/browser_harness/_ipc.py
+++ b/src/browser_harness/_ipc.py
@@ -217,6 +217,15 @@ def expected_token():
 
 
 def cleanup_endpoint(name):  # best-effort; silent if already gone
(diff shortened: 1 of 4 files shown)


Lobster-0429:
Hi, @sauravpanda ,
This is my first contribution.
Could you please review my PR?
Thanks.

Lobster-0429:
Hello @sauravpanda ,
Could  you please review my PR?
I would appreciate your feedback.
Thanks.

sauravpanda:
btw are you a bot?

Lobster-0429:
Hi, @sauravpanda ,
No, I am not a bot.

Lobster-0429:
> btw are you a bot?

How can I contact you?
What do I have to fix now?
Please let me know.

Alezander9:
This PR is far too verbose, the fix only needs a few lines. Done in https://github.com/browser-use/browser-harness/pull/309

Alezander9:
Your account is a bot account, you have lied. Your profile says 7+ years of dev, your account activity goes back 1 month. Your linkedin and portfolio links on profile don't go anywhere. -vouch

PR #292, opened 2026-05-04
fix(daemon): fall back to DevToolsActivePort when BU_CDP_URL returns 404

claytonlin1110:
## Summary
- Handles HTTP 404 from `{BU_CDP_URL}/json/version` the same way profile discovery does: resolve the WebSocket URL from `DevToolsActivePort` under known profile roots.
- Parses host/port with `urlparse` and matches the `DevToolsActivePort` file whose first line equals that port (avoids attaching to the wrong browser when multiple profiles exist).

## Context
Chrome 147+ disables `/json/*` HTTP discovery for the default profile in some setups; `BU_CDP_URL` previously retried for 30s and failed with a generic unreachable error.

## Related
Closes #291

---
## Summary by cubic
Fallback to `DevToolsActivePort` when `BU_CDP_URL` `/json/version` returns 404 to reliably discover the CDP WebSocket. This prevents 30s timeouts and fixes connection failures on Chrome 147+ default profiles.

- **Bug Fixes**
  - Parse host/port from `BU_CDP_URL` and select the `DevToolsActivePort` whose first line matches that port to avoid attaching to the wrong profile.
  - On 404 only, build `ws://{host}:{port}{path}` from `DevToolsActivePort`; keep existing retry logic for other errors and handle trailing slashes.
  - Restore brackets for IPv6 hosts when building the WebSocket URL (e.g., `ws://[::1]:9333/...`) to prevent malformed URLs.

<sup>Written for commit be9316630f9cdbd09d9e9486004e8a7f4eee7d55. Summary will update on new commits.</sup>

PR #300, opened 2026-05-05
fix(run): respect explicit CDP endpoint before cloud auto-bootstrap

ComBba:
## Summary

`BU_AUTOSPAWN` cloud auto-bootstrap currently fires even when the caller
explicitly sets `BU_CDP_URL` or `BU_CDP_WS`, contradicting `install.md:58-59`:

> BU_CDP_WS overrides local Chrome discovery for remote browsers.
> BU_CDP_URL overrides local Chrome discovery with a specific DevTools HTTP endpoint (used for Way 2).

## Why this matters

A caller with `BU_CDP_URL` set for a corporate automation Chrome AND
`BROWSER_USE_API_KEY` set for `profile-use` sync (a common combination —
both env vars get exported once and stay set across unrelated sessions)
currently gets:

1. A billed Browser Use cloud browser provisioned the first time they run
   `browser-harness -c '...'` on a fresh shell.
2. Their explicit endpoint **silently swapped** for the cloud WebSocket
   (see `admin.py:471` smoking gun below).

#181 is an existing report of users already hitting this kind of
`BU_CDP_WS` swap and working around it manually; this PR makes that
workaround unnecessary by honouring the precedence `install.md` promises.

## Reproduction

With `BU_CDP_URL=http://127.0.0.1:9333` plus `BROWSER_USE_API_KEY` and
`BU_AUTOSPAWN=1`, no daemon, and no local Chrome on 9222/9223:

1. `run.py:87-92` does not check explicit endpoints → guard fires.
2. `start_remote_daemon(NAME)` is called → **a billed cloud browser is provisioned**.
3. `admin.py:471` then calls
   `ensure_daemon(env={"BU_CDP_WS": _cdp_ws_from_url(browser["cdpUrl"]), ...})` —
   **overwriting the user's explicit `BU_CDP_URL`** with the cloud WebSocket in the daemon env.

Net: surprise cloud bill plus the explicit endpoint silently replaced.

### Verified against the editable-installed package

| Run | `run.py` state | helper exists | `start_remote_daemon` called |
|---|---|---|---|
| A | upstream main (`origin/main`) | False | **True (bug reproduces)** |
| B | this PR's fix applied | True | **False (correct)** |

Identical env, identical invocation, identical mocks for `daemon_alive` /
`_local_chrome_listening` / `start_remote_daemon`. Only `run.py` differs.

## Fix

Add `_explicit_cdp_configured()` next to `_local_chrome_listening()` and
gate the auto-bootstrap with `not _explicit_cdp_configured()`. Mirrors the
existing `_is_local_chrome_mode` helper at `admin.py:159-161` that already
treats `BU_CDP_WS` as the "not local discovery" signal. #277's
fresh-headless-box behaviour (no explicit endpoint set) is preserved verbatim.

12 LOC change in `run.py` (helper + one guard line + comment refresh).

## Tests

8 new unit tests in `tests/unit/test_run.py`:

- `test_explicit_bu_cdp_url_blocks_cloud_bootstrap` — fix verification (URL)
- `test_explicit_bu_cdp_ws_blocks_cloud_bootstrap` — fix verification (WS)
- `test_both_bu_cdp_url_and_bu_cdp_ws_set_blocks_bootstrap` — caller doubly explicit
- `test_empty_bu_cdp_url_does_not_block_bootstrap` — empty string treated as unset
- `test_explicit_endpoint_does_not_break_daemon_alive_short_circuit` — daemon-alive path preserved
- `test_explicit_endpoint_does_not_break_local_chrome_short_circuit` — local-Chrome path preserved
- `test_explicit_cdp_configured_helper_truthy` — helper input table (URL/WS, IPv6, wss://)
- `test_explicit_cdp_configured_helper_falsy` — helper unset / empty-string

```
$ uv run --with pytest pytest tests/unit/ -q
74 passed in 0.40s
```

Without this patch, exactly 5 of the 8 new tests fail (the ones targeting
the guard + helper); the 3 short-circuit/empty edge cases pass even
unpatched because existing guards already cover those paths — confirms the
new tests isolate the regression cleanly.

## Edge case

If `BU_CDP_URL` is set but unreachable, this PR makes that case fail loudly
via `daemon.py:get_ws_url()`'s 30s probe rather than silently falling back
to a billed cloud browser. If silent fallback was intended, happy to
discuss — but `install.md:58-59` currently reads as an unconditional override.

## Refs

- #266 (original headless auto-bootstrap)
- #277 (`BU_AUTOSPAWN` opt-in by @Alezander9 — same area, didn't address explicit-endpoint precedence)
- #292 (`BU_CDP_URL` `/json/version` 404 fallback by @claytonlin1110)
- #181 (existing user report of manual `BU_CDP_WS` workaround)
- Doc contract: `install.md:58-59`
- Helper precedent: `admin.py:159-161` `_is_local_chrome_mode`
- Smoking gun: `admin.py:471` `start_remote_daemon` overwrites `BU_CDP_WS` in daemon env

### DIFF
--- a/src/browser_harness/run.py
+++ b/src/browser_harness/run.py
@@ -56,6 +56,15 @@ def _local_chrome_listening():
     return False
 
 
+# BU_CDP_URL / BU_CDP_WS are documented to override local Chrome discovery
+# (install.md:58-59), so they must also block cloud auto-bootstrap. Without this
+# guard, start_remote_daemon() in admin.py overwrites BU_CDP_WS in the daemon
+# env with a cloud WebSocket URL, silently replacing the user's explicit endpoint
+# *and* billing them for a cloud browser they never asked for.
+def _explicit_cdp_configured():
+    return bool(os.environ.get("BU_CDP_URL") or os.environ.get("BU_CDP_WS"))
+
+
 def main():
     args = sys.argv[1:]
     if args and args[0] in {"-h", "--help"}:
@@ -83,10 +92,12 @@ def main():
     print_update_banner()
     # Auto-bootstrap a cloud browser is opt-in via BU_AUTOSPAWN — BROWSER_USE_API_KEY alone
     # is not enough, since the key is commonly set for unrelated reasons (profile sync,
-    # cloud API calls, parent agents managing their own session).
+    # cloud API calls, parent agents managing their own session). An explicit BU_CDP_URL
+    # or BU_CDP_WS also blocks the spawn so we honour the precedence install.md promises.
     if (
         not daemon_alive()
         and not _local_chrome_listening()
+        and not _explicit_cdp_configured()
         and os.environ.get("BROWSER_USE_API_KEY")
         and os.environ.get("BU_AUTOSPAWN")
     ):
--- a/tests/unit/test_run.py
+++ b/tests/unit/test_run.py
@@ -29,6 +29,139 @@ def test_cloud_bootstrap_on_headless_server(monkeypatch):
     mock_start.assert_called_once()
 
 
+def test_explicit_bu_cdp_url_blocks_cloud_bootstrap(monkeypatch):
+    """BU_CDP_URL is documented to override local Chrome discovery (install.md:58-59),
+    so it must also block cloud auto-bootstrap. Otherwise start_remote_daemon would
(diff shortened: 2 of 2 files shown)


sauravpanda (approved):

sauravpanda:
Thank you!

PR #307, opened 2026-05-06
fix(daemon): skip unresponsive targets on attach + per-call CDP timeout

decisionm:
## Summary

Two daemon-side bugs cause `browser-harness` to hang silently when Chrome has multiple heavy/sleeping SPA tabs open. The IPC layer's 5s read timeout already lands a clean error in the *calling* client, but the daemon itself stays jammed for every other client — and when the first attach picks a wedged tab, **every** call from then on hits the same dead session.

(Originally this PR also added an IPC-side socket timeout in `helpers.py:_send`, but that bug has already been fixed upstream by `_ipc.connect(NAME, timeout=5.0)`. Rebased to drop the redundant change.)

### 1. `attach_first_page` pinned the daemon to a wedged tab
Picked `targets[0]` unconditionally, ran the four `Domain.enable` calls in parallel, and silently swallowed each timeout. When `targets[0]` was a sleeping tab, busy SPA, or frozen iframe, the daemon happily reported `attached ...` in the log — but every subsequent CDP call against that session hung forever because the renderer was unresponsive.

### 2. `handle()` had no per-call CDP timeout
`await self.cdp.send_raw(method, params, session_id=sid)` with no `wait_for`. One wedged renderer pinned the daemon's event loop forever, starving every other connected client. The IPC `recv` timeout only protected the immediate caller.

## Changes

- **`_try_attach`** (new): attaches to a single target and runs all four `.enable` calls. Returns `None` (and detaches cleanly) if any step fails.
- **`attach_first_page`**: iterates real pages, abandons unresponsive ones, falls back to a fresh `about:blank` if every existing page is unresponsive. Raises if even the fallback fails.
- **`_enable_default_domains`**: gains `strict=False` flag. Initial attach (`strict=True`) returns False on any failure so the caller can abandon the target. `set_session` keeps the existing best-effort behavior since the tab switch has already committed.
- **`handle()`**: wraps `cdp.send_raw` in `asyncio.wait_for` with `BU_CDP_TIMEOUT` (default 45s). Returns a clean error with recovery hint instead of hanging. The stale-session re-attach path is now wrapped in try/except so re-attach failures surface as errors.

`BU_CDP_TIMEOUT` is overridable for legitimately slow workloads (large `captureBeyondViewport` screenshots, long `Runtime.evaluate` scripts).

## Repro

Before fix: open Chrome with heavy SPA tabs (Gmail, Cloudflare dashboard, BBC article, etc.) such that `targets[0]` is unresponsive. Run any helper:

\`\`\`bash
browser-harness -c 'print(page_info())'
\`\`\`

CLI hung indefinitely. Daemon log showed only:
\`\`\`
attached 8F5C7B88... (https://dit.example.com/) session=...
listening on /tmp/bu-default.sock
\`\`\`
…with no clue that the four `.enable` calls had silently timed out behind the scenes.

After fix the iteration is visible and `page_info()` returns in <1s:
\`\`\`
enable Page on D58089BF...: TimeoutError:
enable DOM on D58089BF...: TimeoutError:
enable Runtime on D58089BF...: TimeoutError:
enable Network on D58089BF...: TimeoutError:
target 8F5C7B88... (https://dit.example.com/) unresponsive, trying next
enable Page on 2417ABE3...: TimeoutError:
...
target 784EE4CB... (https://notarycouncil.org/) unresponsive, trying next
...
attached 723889E3... (https://mail.google.com/...) session=4E670D2D...
\`\`\`

## Tests

Three new tests in `tests/unit/test_daemon.py`:

- **`test_attach_first_page_skips_unresponsive_targets`** — wedged target with hanging `Page.enable`; verifies the next live target wins.
- **`test_attach_first_page_creates_blank_when_all_pages_unresponsive`** — every real page wedged; verifies fallback to fresh `about:blank`.
- **`test_handle_times_out_wedged_cdp_call`** — `Runtime.evaluate` hangs forever; verifies `BU_CDP_TIMEOUT` bounds it with a clean error containing the method name and recovery hint.

All 7 pre-existing daemon tests continue to pass (10/10 total).

## Test plan

- [x] Unit tests cover wedged-target skip, all-wedged-fallback, and per-call timeout
- [x] E2E smoke: real Chrome session w/ ~3 wedged SPA tabs → daemon iterates, attaches to live tab, `page_info` returns in <1s
- [x] All pre-existing daemon tests still pass
- [ ] CI green

### DIFF
--- a/src/browser_harness/daemon.py
+++ b/src/browser_harness/daemon.py
@@ -187,23 +187,54 @@ class Daemon:
         self.stop = None  # asyncio.Event, set inside start()
 
     async def attach_first_page(self):
-        """Attach to a real page (or any page). Sets self.session. Returns attached target or None."""
+        """Attach to a real page (or any page). Sets self.session. Returns attached target or None.
+
+        Iterates real pages so a single wedged tab (busy renderer, sleeping tab,
+        unresponsive page) does not pin the daemon to a dead session — previously
+        the daemon picked pages[0] unconditionally, silently swallowed Page.enable
+        timeouts, and every later CDP call against that session hung forever.
+        Falls back to a fresh about:blank if every existing page is unresponsive.
+        """
         targets = (await self.cdp.send_raw("Target.getTargets"))["targetInfos"]
         pages = [t for t in targets if is_real_page(t)]
-        if not pages:
-            # No real pages — create one instead of attaching to omnibox popup
-            tid = (await self.cdp.send_raw("Target.createTarget", {"url": "about:blank"}))["targetId"]
-            log(f"no real pages found, created about:blank ({tid})")
-            pages = [{"targetId": tid, "url": "about:blank", "type": "page"}]
-        self.session = (await self.cdp.send_raw(
-            "Target.attachToTarget", {"targetId": pages[0]["targetId"], "flatten": True}
-        ))["sessionId"]
-        self.target_id = pages[0]["targetId"]
-        log(f"attached {pages[0]['targetId']} ({pages[0].get('url','')[:80]}) session={self.session}")
-        await self._enable_default_domains(self.session)
-        return pages[0]
+        for p in pages:
+            sid = await self._try_attach(p)
+            if sid:
+                self.session = sid
+                self.target_id = p["targetId"]
+                log(f"attached {p['targetId']} ({p.get('url','')[:80]}) session={sid}")
+                return p
+            log(f"target {p['targetId']} ({p.get('url','')[:80]}) unresponsive, trying next")
+        # Either no real pages, or every real page was unresponsive — create a fresh tab.
+        tid = (await self.cdp.send_raw("Target.createTarget", {"url": "about:blank"}))["targetId"]
+        fresh = {"targetId": tid, "url": "about:blank", "type": "page"}
+        sid = await self._try_attach(fresh)
+        if not sid:
+            raise RuntimeError(f"could not attach to any page (created about:blank {tid} but it also failed to enable)")
+        self.session = sid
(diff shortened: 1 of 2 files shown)


decisionm:
@sauravpanda — friendly ping when you have a moment 🙏

This PR fixes two real-world hangs we've been hitting in dm-crm with multi-tab Chrome setups:

1. **Daemon pinned to a wedged tab on initial attach** — `attach_first_page()` used to pick `pages[0]` unconditionally and silently swallow `Page.enable` failures, leaving every later CDP call hanging. Now iterates real pages, abandons unresponsive ones, and falls back to a fresh `about:blank` if every existing page is wedged.

2. **No per-call CDP timeout in `handle()`** — a stalled renderer (sleeping tab, infinite JS, frozen iframe) would block the daemon's event loop forever, starving every other connected client. Adds a per-call bound via `BU_CDP_TIMEOUT` (default 45s).

Just pushed a follow-up commit (`ed94dc2`) addressing the cubic bot's valid feedback: stale-session recovery is now also bounded by `BU_CDP_TIMEOUT` so cumulative per-tab attach attempts can't blow past the contract.

Tests: 11/11 passing (4 new regression tests covering wedged-target skip, all-wedged → about:blank fallback, per-call timeout, and bounded recovery). PR is `MERGEABLE / CLEAN`.

Happy to address any further feedback.

Alezander9:
I cannot reproduce the main issue, and I worry that this adds a lot of code complexity to the attachment point. Try prompting browser-harness to fix itself locally for your use case. Alternatively, perhaps you could more narrowly scope the issue and fix

PR #334, opened 2026-05-10
Document self-modifying Browser Harness workflow

MagMueller:
## Summary
- Add `docs/self-modifying-harness.md` explaining the self-modifying Browser Harness workflow for edge cases.
- Add four concrete edge-case examples: signature/canvas fields, coordinate-only controls, drag-and-drop, and file upload.
- Add `docs/edge-case-benchmark.html`, a local benchmark page covering file upload, drag/drop, signature canvas, and coordinate-only canvas targets.
- Link the docs from `README.md` and `SKILL.md`.

## Validation
- `git diff --check`
- Ran the benchmark in a remote Browser Use Cloud browser by injecting the page into `about:blank`; all four checks passed:
  - file upload: pass
  - drag/drop: pass
  - signature canvas: pass
  - coordinate canvas: pass

## Run notes
- Remote Browser Use Cloud cannot read a local `file://` page directly, so the practical remote path is `Page.setDocumentContent` / injecting the HTML into `about:blank`.
- A long `data:` navigation failed in remote Chrome with `chrome-error://chromewebdata/`; use document injection for this benchmark.
- The installed CLI in this box did not expose `wait_for_element`, so the run used JS polling.
- Signature drawing needed `scrollIntoView()` before measuring coordinates; viewport scroll state matters for canvas tests.
- Repeated page-context snippets should be wrapped in IIFEs to avoid global `const` redeclaration errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ See [agent-workspace/domain-skills/](agent-workspace/domain-skills/) for example
 Read [Self-modifying Browser Harness](docs/self-modifying-harness.md) for the
 core workflow: when a task hits an edge case, the agent patches the harness and
 retries. The guide walks through file uploads, drag-and-drop, signature/canvas
-drawing, and coordinate-only controls.
+drawing, coordinate-only controls, and a local edge-case benchmark page.
 
 ## Free Browser Use Cloud browsers
 
--- a/README.md
+++ b/README.md
@@ -38,6 +38,11 @@ Click Allow when the per-attach popup appears (Chrome 144+):
 
 See [agent-workspace/domain-skills/](agent-workspace/domain-skills/) for example tasks.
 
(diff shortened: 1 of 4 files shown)


MagMueller:
Worker verification pass (post-bench-docs):

- Local worktree HEAD matches `origin/codex/self-modifying-harness-docs` at `2101f01`.
- Files in place: `docs/self-modifying-harness.md` (8.2 KB), `docs/edge-case-benchmark.html` (8.7 KB), `README.md` link at L41, `SKILL.md` pointer at L84.
- `git diff --check`: clean.
- HTML sanity: well-formed `<!doctype html>` … `</html>`, no stray markers.
- Checks: CodeQL ✅, GitGuardian ✅, skill-review ✅ (cubic still running at comment time).

Ready for review/merge from the docs side. Edge-case benchmark notes are owned by another worker.

PR #393, opened 2026-05-27
fix: close_tab() recovers session state when closing the currently attached tab

### LINKED ISSUE #379: close_tab() on the current tab leaves dangling session state
## Context

Follow-up from #337, which added `close_tab(target=None)` to `src/browser_harness/helpers.py`. Merged with this known sharp edge.

## Problem

`switch_tab()` registers the active tab/session in the daemon via:

```python
_send({"meta": "set_session", "session_id": sid, "target_id": target_id})
```

`close_tab()` calls `Target.closeTarget` directly and does **not** clear or update that state. So when you call `close_tab()` with no args (closing the currently-attached tab), the daemon still believes that now-dead `targetId` / `sessionId` is active.

Subsequent calls that depend on the attached session — `js()`, `page_info()`, `current_tab()`, `_mark_tab()`, etc. — will hit a dead session and fail or misbehave until the caller manually `switch_tab()`s to another tab.

The bulk-cleanup example in the PR description sidesteps this by only closing non-current tabs, but the no-arg `close_tab()` shape is the obvious footgun.

## Repro

```python
close_tab()           # closes the attached tab
page_info()           # boom — session points at a closed target
```

## Suggested fix

When `close_tab()` is closing the currently-attached tab, fall back to another real tab before issuing `Target.closeTarget`:

- If `list_tabs(include_chrome=False)` has another tab, `switch_tab()` to it first, then close the old `targetId`.
- If no other real tab exists, either (a) open a fresh `about:blank` and switch to it, or (b) clear the daemon session and document that the caller must `switch_tab()` / `new_tab()` before further calls.

Option (a) matches the spirit of `ensure_real_tab()` and keeps the harness in a usable state. Option (b) is simpler but pushes the burden onto callers.

Either way, add a short note to the docstring.

## Scope

- `src/browser_harness/helpers.py` — update `close_tab()`.
- Worth a small test: open two tabs, close current, assert that subsequent `current_tab()` / `js()` works.

mvanhorn:
## Summary

`close_tab()` now switches the harness to a surviving real tab (or a fresh `about:blank`) before closing the currently attached tab, so the daemon session never points at a dead target.

## Why this matters

Before this change, the no-arg form of `close_tab()` issued `Target.closeTarget` against the attached `targetId` without updating the daemon session that `switch_tab()` previously registered. After `close_tab()` the daemon still believed the now-dead `targetId`/`sessionId` was active, so the next call into `js()`, `page_info()`, `current_tab()`, or `_mark_tab()` landed on a dead session and failed until the caller manually `switch_tab()`'d.

Repro:
```python
close_tab()
page_info()  # fails: not attached to live tab
```

Maintainer @sauravpanda filed #379 with a clear cause + Option (a) fix sketch.

## Changes

- `close_tab(target=None)` now detects whether the target equals the currently attached tab. If so, it picks a surviving real tab via `list_tabs(include_chrome=False)` and `switch_tab()`s to it before closing; if no other real tab exists, it opens a fresh `about:blank` first. The fast path for explicit non-current targets is unchanged.
- Error path: if `Target.closeTarget` fails after the switch, the original session is restored rather than left half-migrated.

## Testing

`pytest tests/unit/test_helpers.py` — 22/22 pass. Added 4 new test cases:
- Happy path: two tabs, no-arg close, current tab + daemon session land on the survivor.
- Single-tab fallback: about:blank created, session points at it, `page_info()` succeeds.
- Explicit non-current close: current tab + session unchanged.
- Error path: simulated close failure restores original session intact.

Fixes #379

---
## Summary by cubic
Prevents the daemon session from pointing at a dead tab when closing the attached tab. `close_tab()` switches to a surviving real tab (or a fresh about:blank) before closing and performs best‑effort rollback and cleanup on failure. Fixes #379.

- Closing a non‑current tab leaves the session unchanged.
- On failure: restore the original session, close any temporary about:blank tab, and treat `Target.closeTarget` responses with `success: false` as failures; rollback failures never mask the original close error.
- Adds unit tests for survivor switch, blank fallback, explicit non‑current close, single‑tab failure cleanup, and simultaneous close+rollback failures.

<sup>Written for commit efa1e7690d0a4f49878cb2a463f2f7133fbf4db6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/393?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mvanhorn:
Good catch by cubic, the leak was real. Fixed in f33ea09: the single-tab fallback tab id is tracked and best-effort closed when `Target.closeTarget` fails, before re-raising, and `success: false` responses are now treated as failures too. Added fake-CDP tests for both failure modes; 24 unit tests pass.

mvanhorn:
Fixed. `close_tab()` rollback is now best-effort, so a rollback failure can no longer mask the original close error or skip the fallback cleanup. Added regression coverage for the case where close and rollback both fail.

25 tests pass in the unit suite.

PR #402, opened 2026-06-02
Add BH_NO_ACTIVATE flag to suppress focus-steal (switch_tab + new_tab + daemon)

undeemed:
## Problem

`new_tab()` → `switch_tab()` → `cdp("Target.activateTarget", ...)` raises the tab/window to the OS **foreground on every call**. When driving a *visible* browser (e.g. an isolated debug Chrome the user is watching), this steals focus on every single action — the window keeps popping to the front.

## Change

Gate the `Target.activateTarget` call behind an opt-in env flag:

```python
if os.environ.get("BH_NO_ACTIVATE") != "1":
    cdp("Target.activateTarget", targetId=target_id)
```

- **Default behavior unchanged** — activation still happens unless `BH_NO_ACTIVATE=1` is set.
- Routing is unaffected: `set_session` (right below) is what makes `cdp()`/`js()` default to the tab; activation is purely cosmetic OS focus, and the 🐴 title marker already shows which tab is controlled.
- The flag is read client-side in `switch_tab`, so callers just prefix runs with `BH_NO_ACTIVATE=1` — no daemon restart needed.

## Use case

Automated/background drivers attached to a visible isolated browser, where focus-stealing on each `new_tab` is disruptive.

---
## Summary by cubic
Added `BH_NO_ACTIVATE` to suppress focus-steal: `switch_tab()` skips `Target.activateTarget`, and `new_tab()` and the daemon pass `background:true` to `Target.createTarget` so tabs open without raising the window (verified on macOS). Defaults are unchanged; when the flag is set the daemon anchors a real `data:` tab to keep backgrounding reliable, and you can override per call via `new_tab(background=...)` or `switch_tab(activate=...)`.

<sup>Written for commit 929b4dc9e4b48c0df0ddbf58a713d49aaa7f1eab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/402?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -79,6 +79,7 @@ Cloud profile cookie sync reference: https://github.com/browser-use/browser-harn
 - Clicking: screenshot -> read pixel -> `click_at_xy(x, y)` -> screenshot again.
 - After navigation, call `wait_for_load()`.
 - If the current tab is stale or internal, call `ensure_real_tab()`.
+- Driving a *visible* browser the user is watching? Set `BH_NO_ACTIVATE=1` so `new_tab()`/`switch_tab()` don't raise the window and steal OS focus. Per call: `new_tab(url, background=True|False)` or `switch_tab(target, activate=True|False)`. The daemon auto-anchors a tab, so no manual sentinel needed.
 - Use `js(...)` for DOM inspection or extraction when coordinates are the wrong tool.
 - Login walls: stop and ask. Exception: use available SSO automatically when Chrome is already signed in; still stop for passwords, MFA, consent, or ambiguous account choice.
 - Raw CDP is available with `cdp("Domain.method", ...)`.
--- a/install.md
+++ b/install.md
@@ -83,4 +83,6 @@ browser-harness --update -y
 browser-harness telemetry disable
 ```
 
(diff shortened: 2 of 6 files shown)


undeemed:
Closing — opened prematurely, will revisit on my own terms. Apologies for the noise.

aiba:
Wait this seems super useful.  Can we reopen this so it can be merged?

undeemed:
Reopening — closed this early in June, complete now. `switch_tab`'s gate alone left the `Target.createTarget` raise, so I added the `background:true` half for `new_tab` + daemon bootstrap (thanks @aiba for flagging that path), plus a per-call `new_tab(url, background=…)` override and a daemon keepalive anchor so it works without a user-managed sentinel (the #469 follow-up). Documented the flag, added tests (incl. an activate-skip regression), verified e2e on Chrome for Testing. Rebased onto main. Overlaps #469 (same core fix) — can be the single PR to review, no preference which lands.

undeemed:
Only verified the focus-steal fix on macOS. The `background:true` + skip-`activateTarget` path is unconditional Chromium C++ (no mac-specific code), so it should hold cross-platform — but if anyone's on Windows or Linux, a quick confirm that `new_tab()` under `BH_NO_ACTIVATE=1` doesn't raise the window would be appreciated.

Issue #352, opened 2026-05-13
Design proposal: make tab/session lifecycle recovery explicit


honor2030:
## Before submitting

- [x] I searched existing issues and PRs.
- [x] This is a feature/change request, not a single bug report.

Related open PRs I found while checking overlap:

- #182 — `switch_tab()` can hang when the previously attached session is stale/dead.
- #208 — startup attach can pick a discarded tab whose renderer never responds.
- #347 — `no close frame received or sent` should trigger re-attach instead of surfacing as a hard failure.

Those PRs all look useful individually, but they also point at a shared lifecycle problem around “what is the currently attached tab/session, and how do helpers recover when it is no longer live?”

## Problem

`browser-harness` currently handles stale tab/session states in several focused places:

- `attach_first_page()` picks a page target and enables default domains.
- `set_session` updates `self.session` / `self.target_id` after `switch_tab()` and `new_tab()`.
- default domains are re-enabled after session changes so helpers like `wait_for_network_idle()` keep working.
- `current_tab` has a `not_attached` path when `target_id` is missing.
- stale-session recovery in `handle()` currently depends on specific error text such as `Session with given id not found` and, in #347, `no close frame`.
- cosmetic/best-effort work such as tab-title unmarking can still become a blocking operation if it talks to a dead renderer.

The result is that each new CDP edge case tends to be fixed locally with another timeout/string-match/reattach patch. That is pragmatic, but it makes it hard to know the intended invariant:

> Before a helper performs a renderer-bound operation, what guarantees do we have about the current target/session? If those guarantees fail, where is recovery supposed to happen?

This is mostly visible in stale-target cases: closed tabs, discarded tabs, Chrome restarts, expired sessions, or WebSocket/session close paths.

## Proposal

Add a small explicit internal contract for tab/session lifecycle recovery, without changing the public helper API initially.

Possible shape:

1. Define the daemon’s internal session states / failure classes, e.g.
   - `no_session`
   - `attached`
   - `target_missing`
   - `session_stale`
   - `renderer_unresponsive`
   - `browser_disconnected`
2. Centralize recovery in one small helper such as `_ensure_live_session()` / `_recover_session()` / `_with_live_session()` that is used by daemon request handling and session-changing paths.
3. Treat cosmetic operations, such as removing the controlled-tab marker from the old tab, as explicitly best-effort and bounded by a short timeout.
4. Keep the first implementation slice small and testable:
   - no behavior change for healthy tabs;
[...]

PR #417, opened 2026-06-10
Fix cdp() params-dict form and new_tab() blank-tab leak

pA1nD:
## Problem

`SKILL.md` line 93 documents raw CDP as:

> Raw CDP for anything helpers don't cover: `cdp("Domain.method", params)`.

But the implementation is `cdp(method, session_id=None, **params)` — so the documented dict form puts the params dict into `session_id` and sends empty params. Every agent following the skill doc hits this:

```python
cdp("Target.closeTarget", {"targetId": tid})
# RuntimeError: {'code': -32600, 'message': "Message may have string 'sessionId' property"}
# (or, depending on method: "Failed to deserialize params.targetId - BINDINGS: mandatory field missing")
```

A practical consequence: tab-close calls written in the documented style fail, so agents quietly leak tabs.

## Fix

`cdp()` accepts both the documented dict form and the existing kwargs form (kwargs win on key collisions), and `session_id` becomes keyword-only:

```python
def cdp(method, params=None, *, session_id=None, **kwargs):
```

Compatibility: every call site in `src/`, `agent-workspace/`, `interaction-skills/` and `agent-workspace/domain-skills/` already passes `session_id` by keyword (grepped; none positional), so this is non-breaking for the repo and for code following either documented style.

## Related: `new_tab()` leaks its about:blank tab on failure

`new_tab()` deliberately creates `about:blank` first and navigates after (to avoid the createTarget/attach race). If `switch_tab`/`goto_url` then throws, the blank tab is orphaned — over many sessions these accumulate visibly. Now the blank tab is closed before the exception re-raises.

## Tests

5 new unit tests in `tests/unit/test_helpers.py` (kwargs form, dict form, merge precedence, keyword `session_id`, new_tab close-on-fail). Full suite: 112 passed.

---
## Summary by cubic
Fixes `cdp()` to accept the documented params-dict form and stops `new_tab()` from leaking the temporary about:blank tab on failures. Aligns behavior with `SKILL.md` and prevents orphaned tabs.

- **Bug Fixes**
  - `cdp()` now accepts both dict and kwargs forms; kwargs win on collisions, and `session_id` is keyword-only. This resolves failures like `Target.closeTarget` when called as `cdp("Domain.method", {...})`. No breaking change.
  - `new_tab()` closes the about:blank tab if attach or navigation fails, then re-raises, preventing tab leaks.

<sup>Written for commit 2d5e40da873b53f8093d6137601f22d1ec6a0803. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/417?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/browser_harness/helpers.py
+++ b/src/browser_harness/helpers.py
@@ -49,9 +49,10 @@ def _send(req):
     return r
 
 
-def cdp(method, session_id=None, **params):
-    """Raw CDP. cdp('Page.navigate', url='...'), cdp('DOM.getDocument', depth=-1)."""
-    return _send({"method": method, "params": params, "session_id": session_id}).get("result", {})
+def cdp(method, params=None, *, session_id=None, **kwargs):
+    """Raw CDP. Kwargs and a params dict both work:
+    cdp('Page.navigate', url='...'), cdp('Target.closeTarget', {'targetId': tid})."""
+    return _send({"method": method, "params": {**(params or {}), **kwargs}, "session_id": session_id}).get("result", {})
 
 
 def drain_events():  return _send({"meta": "drain_events"})["events"]
@@ -319,9 +320,15 @@ def new_tab(url="about:blank"):
     # attach, so the brief about:blank is "complete" by the time the caller
     # polls and wait_for_load() returns before navigation actually starts.
     tid = cdp("Target.createTarget", url="about:blank")["targetId"]
-    switch_tab(tid)
-    if url != "about:blank":
-        goto_url(url)
+    try:
+        switch_tab(tid)
+        if url != "about:blank":
+            goto_url(url)
+    except Exception:
+        # a failed attach/navigation must not leak the blank tab
+        try: cdp("Target.closeTarget", targetId=tid)
+        except Exception: pass
+        raise
     return tid
 
 def close_tab(target=None):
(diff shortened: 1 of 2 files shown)


pA1nD:
linking some related issues i found while digging: this sits on the same tab/session lifecycle seam as #379 and #352, might help triage.

one honest caveat on the new_tab guard: if goto_url fails after switch_tab already ran, we close a tab the daemon still considers current. that is exactly the dangling session state #379 describes. i figured leaking a session pointer (recovers on the next call) beats leaking a tab (sits there forever, rent free), but if #393 lands first this cleanup could route through that instead.

for what it's worth, while testing this i found four orphaned about:blank tabs in my own browser. at least they had each other.

Kastan97 (commented):

**What this does:** Two fixes: (1) cdp() now accepts both keyword args and an explicit params=dict; (2) new_tab() cleans up blank targets on attach failure.

Good to merge once CI passes.

PR #419, opened 2026-06-10
docs: warn that press_key/fill_input double-inserts characters on newer Chrome

blaircullen:
## What

Adds a field-tested gotcha to `SKILL.md`: `press_key` (and therefore `fill_input`, which calls it per character) **double-inserts every character on newer Chrome**.

## Repro

```py
new_tab("about:blank")
js("document.body.innerHTML='<input id=t>';document.querySelector('#t').focus();")
for ch in "abc": press_key(ch)
js("(()=>document.querySelector('#t').value)()")   # -> "aabbcc"   (expected "abc")
```

Confirmed on **HeadlessChrome 149.0.0.0**. Real-world symptom: a login filled via `fill_input` sent `mm33mmeeddiiaa` for `m3media` and the site returned **"Invalid username or password"** — looked like a credential problem, was actually doubled input.

## Root cause

`press_key` dispatches `keyDown` **with** a `text` field *and* a separate `char` event:

```py
cdp("Input.dispatchKeyEvent", type="keyDown", **base, **({"text": text} if text else {}))
if text and len(text) == 1:
    cdp("Input.dispatchKeyEvent", type="char", text=text, ...)
```

On newer Chrome the `keyDown` carrying `text` already inserts the character, so the explicit `char` event inserts it a second time. (On older builds `keyDown(text)` apparently didn't auto-insert, so the pair produced a single insert — hence it worked before.)

## This PR

Documentation only — adds the symptom, root cause, and a native-value-setter workaround for controlled inputs.

## Suggested real fix (not in this PR)

Make `press_key` insert exactly once on all Chrome versions — either drop `text` from the `keyDown` and let the `char` event do the insert, or send `rawKeyDown` + `char`. Happy to follow up with the code change + a regression test if you'd prefer it in the same PR.


Kastan97 (commented):

**What this does:** Adds a one-line warning in SKILL.md that press_key/fill_input double-inserts characters on newer Chrome. Useful operational guidance.

Good to merge once CI passes.

PR #508, opened 2026-07-12
fix: per-call CDP timeout + auto-reconnect on dropped websocket

Rselectronic:
Two failure modes that made the harness feel slow and flaky in daily use.

## #1 — Every helper→daemon call was hard-capped at 5s
`ipc.connect()` sets the connect **and** read timeout to a single value, and `_send` passed `5.0`. So any single CDP command that legitimately took longer — screenshots on a large/retina page, heavy `DOM.getDocument`, awaited promises, slow navigations — raised `socket.timeout` and killed the script.

Fix: split the two budgets. Keep a short 5s **connect** budget; add a separate **command read** budget (default 30s, screenshots 60s), tunable via `BH_CMD_TIMEOUT` and per-call `cdp(_timeout=)` / `js(timeout=)`.

## #2 — Daemon never rebuilt a dropped websocket
The CDP client was built once in `start()`. If the websocket dropped (Chrome closed/restarted, remote hiccup), every subsequent call returned an error until a manual `browser-harness --reload` — a zombie daemon that still answered `ping`.

Fix: `_conn_dead()` detection + a locked `_reconnect()` that rebuilds the client and retries the failing call once. For a local Chrome, `get_ws_url()` re-resolves the live websocket, so it also recovers from a Chrome restart on the same port. Only if the rebuild itself fails (e.g. a remote `BU_CDP_WS` whose endpoint is gone) is the error surfaced.

## Verification
- **#1 live:** an 8s awaited promise now returns (was hard-fail at 5s); `BH_CMD_TIMEOUT=2` fails fast at 2s.
- **#2 live:** killed Chrome mid-session, restarted it, next call self-healed (`Connection lost` → `reconnecting` → `attached` → OK).
- New deterministic coverage in `tests/unit/test_reconnect.py` (5 tests). Full suite: **112 passed**.

## Notes
- Tradeoff: a genuinely hung Chrome now waits the command budget (30s default) before failing instead of 5s. Intended — 5s was too short for real work — and tunable down via `BH_CMD_TIMEOUT`.
- `send_raw` still has no server-side timeout; the client read budget is the only bound. Out of scope here.

---
## Summary by cubic
Split the connect vs command timeouts to remove the 5s hard cap, and added automatic reconnect when the CDP websocket drops. Long CDP calls now succeed, and the daemon self-heals after Chrome restarts or transient drops.

- **Bug Fixes**
  - Separate timeouts: keep a 5s connect budget; add a per-call command read budget (30s default, screenshots 60s). Tunable via `BH_CMD_TIMEOUT`; override per call with `cdp(_timeout=)` or `js(timeout=)`.
  - Auto-reconnect: detect a dropped websocket, rebuild the CDP client under a lock, and retry the call once. For local Chrome, `get_ws_url()` re-resolves the live socket; if a remote endpoint is gone, the error is returned.

<sup>Written for commit 547dc0e1571b791e1741b31573824985c8d614a8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/508?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,8 @@
 # Copy to .env — auto-loaded. Only needed for remote browsers.
 BROWSER_USE_API_KEY=bu_your_key_here
+
+# Optional. Per-call read budget (seconds) for every CDP command the helpers
+# send to the daemon. Default 30 (screenshots 60). Raise it for pages with very
+# slow navigations or long awaited promises; lower it to fail fast. Individual
+# calls can also override via cdp(..., _timeout=N) / js(..., timeout=N).
+# BH_CMD_TIMEOUT=30
--- a/interaction-skills/connection.md
+++ b/interaction-skills/connection.md
@@ -6,6 +6,24 @@ When Chrome opens fresh, the only CDP `type: "page"` targets are `chrome://inspe
 
 The daemon's `attach_first_page()` handles this by creating an `about:blank` tab when no real pages exist. If you still end up on an invisible tab, use `switch_tab()` which calls `Target.activateTarget` to bring the tab to front.
 
+## Auto-reconnect
+
+If the CDP websocket drops mid-session (Chrome closed/restarted, a remote
+endpoint hiccup), the daemon rebuilds the connection and retries the failing
+call once — for a local Chrome it re-resolves the live websocket via
+`get_ws_url()`, so it also recovers from a Chrome restart on the same port. The
+call the caller made succeeds transparently; a follow-up call re-attaches any
+tab-specific session. Only if the rebuild itself fails (e.g. a remote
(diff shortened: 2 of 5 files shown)


lylepratt:
Also running into this problem. This hardcoded timeout should be configurable.

PR #523, opened 2026-07-14
fix: avoid retrying user JavaScript errors

ShiroKSH:
## Summary

- Retry a JavaScript snippet only for Chrome's exact top-level-return parser error.
- Avoid a second evaluation after user code throws an error containing the same text.
- Add regressions for user-thrown `Error` and `SyntaxError` values.

## Root cause

The retry decision used a substring of a flattened runtime error. User code could throw that text after changing page state, causing the harness to execute the expression twice.

## Checks

- `uv run --with pytest pytest tests/integration/test_js.py tests/unit/test_helpers.py -q`
- `git diff --check`

---
## Summary by cubic
Only retry JavaScript evaluation for Chrome’s exact top‑level return SyntaxError. This prevents double execution and side effects when user code throws errors that mention “Illegal return statement”.

- **Bug Fixes**
  - Detect Chrome’s top‑level return via `result` fields and raise `_IllegalTopLevelReturn`; `helpers.js` retries once by wrapping the expression.
  - Remove substring-based matching; do not retry on user‑thrown `Error`/`SyntaxError` that include “Illegal return statement”.
  - Add tests covering Chrome’s error path and regressions for user‑thrown errors.

<sup>Written for commit 182783eb986bdf7d1a798bf9b111d9dcb89ddabb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/523?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/browser_harness/helpers.py
+++ b/src/browser_harness/helpers.py
@@ -91,6 +91,10 @@ def _decode_unserializable_js_value(value):
     return value
 
 
+class _IllegalTopLevelReturn(RuntimeError):
+    pass
+
+
 def _runtime_value(response, expression):
     result = response.get("result", {})
     details = response.get("exceptionDetails")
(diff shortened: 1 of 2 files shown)


Kastan97 (commented):

**What this does:** Correctly identifies `Illegal return statement` via CDP's `className` (`SyntaxError`) instead of fragile string matching on the error message. Prevents retrying genuine user `throw` errors that happen to mention 'Illegal return' in their message.

Good to merge once CI passes.

ShiroKSH:
Hi maintainers, this PR has been open for over two weeks. When you have a moment, I would appreciate a review. Thank you.

PR #525, opened 2026-07-14
Add X DM reply-batch and spreadsheet paste recipes

rishwajeet:
## What changed

- Added an X direct-message recipe covering authenticated session recovery, stable DM selectors, request acceptance, composer verification, bounded batches, and idempotent reruns.
- Added a clipboard recipe for spreadsheet ranges that ignore synthetic Cmd+V.
- Added the clipboard recipe to the Interaction Skills index. The X recipe is discovered through the existing domain-skill routing when `BH_DOMAIN_SKILLS=1`.

## Why

X appends timestamps inside message wrappers. Exact-text verification can therefore miss a successful send and resend the same DM. Google Sheets can also show the intended range as selected while ignoring a synthetic paste.

## Impact

Future reply batches can detect existing outgoing drafts safely, accept only the requests that actually need replies, and verify each send in-thread. Tracker updates have a documented paste fallback plus an explicit live-read verification step.

## Checks

- Compiled all six Python code blocks with Python 3.13 via `uv run python`.
- Ran `git diff --check`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ build/
 dist/
 .idea/
 .claude/
+.DS_Store
--- a/SKILL.md
+++ b/SKILL.md
@@ -98,6 +98,7 @@ Cloud profile cookie sync reference: https://github.com/browser-use/browser-harn
 If you get stuck on a browser mechanic, check https://github.com/browser-use/browser-harness/tree/main/interaction-skills.
 
 - connection.md
+- clipboard.md
 - cookies.md
 - cross-origin-iframes.md
 - dialogs.md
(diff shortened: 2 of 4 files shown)


Kastan97 (commented):

**What this does:** Adds an X DM reply-batch recipe (155 lines) covering authenticated session recovery, stable DM selectors, request acceptance, composer verification, bounded batches, and idempotent reruns. Also adds a clipboard recipe (60 lines) for spreadsheet ranges that ignore synthetic Cmd+V, and wires it into the Interaction Skills index.

**Assessment:** Well-structured documentation additions. The clipboard recipe addresses a real friction point (Google Sheets ignoring synthetic paste). The X DM recipe includes practical guardrails like bounded batches and idempotency checks. 217 lines total of pure documentation.

Good to merge.

Issue #554, opened 2026-07-24
Cold daemon against a windowless Chrome hangs 45s, then misreports as an unclicked "Allow remote debugging?" popup


illmethinks0:
## Summary

When the daemon starts cold against a Chrome that has **zero open windows** (e.g. Chrome launched with `--no-startup-window`), the browser-endpoint WS handshake stalls for the full 45s timeout and then fails with:

```
RuntimeError: permission-blocked: wait for the user to click Allow in the Chrome permission popup before retrying.
```

No popup is ever shown, because Chrome has no window to render one in. The user is told to click something that does not exist, and there is nothing in the message pointing at the actual cause.

Opening any Chrome window makes the identical cold start succeed immediately.

## Environment

- browser-harness 0.1.7 (git, editable install via uv)
- Google Chrome 150.0.7871.184
- macOS 26.5.2 (Apple Silicon)
- Local Chrome mode (no `BU_CDP_WS` / `BU_CDP_URL`)
- `devtools.remote_debugging.user-enabled = True` in Local State, `DevToolsActivePort` live on 9222
- `remote_debugging_user_enabled()` returns `True`

## Reproduction

Chrome running with `--no-startup-window` and no windows open:

```bash
osascript -e 'tell application "System Events" to tell process "Google Chrome" to return count of windows'
# 0

pkill -f browser_harness.daemon
browser-harness <<'PY'
print(page_info())
PY
```

**Actual** (deterministic, reproduced on consecutive runs):

```
browser-harness: Chrome is asking "Allow remote debugging?" — click Allow to continue.
RuntimeError: permission-blocked: wait for the user to click Allow in the Chrome permission popup before retrying.
```

Daemon log:

```
connecting to ws://127.0.0.1:9222/devtools/browser/dddfa87e-...
handshake-wait: if Chrome shows an 'Allow remote debugging?' popup, click Allow
fatal: permission-blocked: Chrome's 'Allow remote debugging?' popup was not accepted within 45s -- wait for the user to click Allow, then retry
```

Now open a single Chrome window and repeat the same cold start:

```bash
# open any window, then
pkill -f browser_harness.daemon
browser-harness <<'PY'
print(page_info())
PY
# {'url': 'about:blank', 'title': '', 'w': 1440, 'h': 713, ...}
```

Succeeds. Only variable changed is window count 0 -> 1.

## Why the message is misleading

[...]

PR #555, opened 2026-07-24
Narrow skill description to match When Not to Use

illmethinks0:
## What

The `SKILL.md` frontmatter description still reads:

> Always use browser-harness for any web interaction: automation, scraping, testing, or site/app work.

0.1.7 added a **When Not to Use** section that says the opposite for the simple cases - prefer `curl` for a plain HTTP fetch of public content. The description was never updated to match. `git log -L 3,3:SKILL.md` shows it unchanged apart from the browser-use -> browser-harness rename.

## Why it matters

In Claude Code and Codex the frontmatter `description` is the always-loaded field, and it is what decides whether the skill gets invoked at all. The body only loads once the skill fires. So the always-visible text tells agents to always use the browser for any web interaction, while the correction sits in the half that may never be read.

The practical effect is agents opening a browser for tasks the skill itself says should use `curl`.

## Change

One line. Keeps trigger coverage for genuine browser work (interaction, logged-in sessions, JS-rendered or bot-protected pages) and drops the claim over plain fetches.

Related to #554 but independent - that one is daemon connection classification, this is skill metadata.

---
## Summary by cubic
Updates `SKILL.md`'s description so agents stop assuming `browser-harness` should drive every web interaction. It now reserves browser use for real-browser work (clicking, typing, navigation, logged-in, JS-rendered or bot-protected pages) and points plain public fetches at `curl`, matching the existing When Not to Use section; the description is the always-loaded field, so the old text made agents launch browsers unnecessarily.

<sup>Written for commit 4e519864235692de16b168672158071b2ca58cb3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/555?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: browser-harness
-description: "Always use browser-harness for any web interaction: automation, scraping, testing, or site/app work."
+description: "Control a real browser via CDP: clicking, typing, navigation, logged-in sessions, JS-rendered or bot-protected pages. Not for plain HTTP fetches of public content - use curl for those."
 ---
 
 # browser-harness
--- a/tests/unit/test_skill.py
+++ b/tests/unit/test_skill.py
@@ -31,5 +31,5 @@ def test_packaged_skill_frontmatter_is_valid_simple_yaml():
 
     assert metadata == {
         "name": "browser-harness",
-        "description": "Always use browser-harness for any web interaction: automation, scraping, testing, or site/app work.",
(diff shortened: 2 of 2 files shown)


Kastan97 (commented):

**What this does:** Narrows skill description to align with the existing When Not to Use section. Matching docs to intended scope.

Good to merge once CI passes.

PR #585, opened 2026-08-06
macOS: use `open` instead of AppleScript to reach Chrome

atnickallen:
## The problem

The docs currently recommend AppleScript over `open` when Chrome is already running, so it reuses the current profile instead of going through the profile picker (`SKILL.md`, `install.md`, `interaction-skills/connection.md`):

```bash
osascript -e 'tell application "Google Chrome" to activate' \
          -e 'tell application "Google Chrome" to open location "chrome://inspect/#remote-debugging"'
```

`tell application` is an Apple Event, and macOS gates Apple Events behind an Automation consent dialog — the square *"&lt;app&gt; wants access to control Google Chrome"* popup with Don't Allow / Allow.

The part that makes this worth changing: **that grant is keyed to the exact binary path of the calling process.** Agents driving the harness typically run under Homebrew's node, which lives at `/opt/homebrew/Cellar/node/<version>/bin/node` — the version is *in the path*. Every `brew upgrade node` moves the binary and silently invalidates the grant.

From the user's side it looks unfixable. They click Allow, it works, and some days later the popup is back with no apparent cause. Verified in a TCC dump, where the Chrome grant existed for the old node but the agent had since moved to a newer one:

```
/opt/homebrew/Cellar/node/25.4.0/bin/node -> com.google.Chrome  auth=2
/opt/homebrew/Cellar/node/26.5.0/bin/node -> (no Chrome entry)
```

Anyone can reproduce the mechanism without the harness:

```bash
osascript -e 'tell application "Google Chrome" to activate'   # prompts, keyed to the caller's path
open -a "Google Chrome"                                        # never prompts
```

## The change

`open` is not an Apple Event and needs no grant, so it survives node upgrades.

The tradeoff is real but much cheaper: `open` may route through the profile picker when Chrome is **not** already running. The surrounding docs already tell the agent how to handle the picker, and a one-time picker is better than a permission dialog that returns forever.

Docs-only. Updates the three places that recommended AppleScript:

- `SKILL.md` — the gotcha bullet
- `install.md` — the escalation step and the checklist line
- `interaction-skills/connection.md` — the "Bringing Chrome to front" snippet

The AppleScript in `interaction-skills/tabs.md` is deliberately left alone: reading the visible tab-strip order genuinely needs Apple Events, and it is not on the startup path, so it does not fire during routine attach.

---
## Summary by cubic
Switch macOS docs to use `open` instead of AppleScript to reach Chrome, removing the recurring Automation consent popup tied to Homebrew `node` path changes. Updates `SKILL.md`, `install.md`, and `interaction-skills/connection.md`; a profile picker may appear if Chrome isn’t already running.

- **Migration**
  - Use: `open -a "Google Chrome" "chrome://inspect/#remote-debugging"`
  - Avoid AppleScript `tell application` on startup/attach paths; AppleScript in `interaction-skills/tabs.md` stays for tab-strip reads.

<sup>Written for commit a38924b41e958acf8139113a400ef010ed964288. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/585?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


atnickallen:
Closing this — withdrawing it as filed.

I opened it while diagnosing a recurring "Allow remote debugging?" popup, and I attributed that popup to this AppleScript path. That attribution was wrong. The dialog was Chrome's own per-connection consent prompt, which Chrome shows for debugging enabled via the `chrome://inspect` toggle; it had nothing to do with Apple Events. The real fix was launching Chrome with a command-line `--remote-debugging-port` on a non-default user-data-dir, which never prompts.

The technical claim here is still accurate on its own terms — `tell application` does require a TCC Automation grant, and that grant is keyed to the caller's full binary path, so a Homebrew node upgrade does invalidate it. But I framed the PR around a problem it does not solve, and I would rather withdraw it than leave a maintainer to evaluate a change sold on a faulty premise.

If the underlying point is useful to you, it would be better filed as a plain note about Apple Events and TCC, without the causal story I attached to it. Sorry for the noise.

PR #600, opened 2026-08-13
fix: open remote debugging in the running Chromium browser

### LINKED ISSUE #425: Multi-browser setups (Dia/Arc + Chrome installed): stale Chrome DevToolsActivePort shadows the active browser; recovery flow force-opens Google Chrome
## Environment

- macOS 15 (Darwin 25.4), browser-harness 0.1.0 (git, current main @ 6d20866)
- Primary browser: **Dia** (Chromium-based), remote debugging enabled via the `chrome://inspect/#remote-debugging` checkbox, listening on `127.0.0.1:9222`. Note: with the new consent model, `/json/version` on that port returns **404**, and a per-connection "Allow remote debugging" dialog is shown in the browser.
- Google Chrome is installed but was **not running**. A stale `DevToolsActivePort` file (pointing at a dead ephemeral port) was left behind in `~/Library/Application Support/Google/Chrome` from an earlier Chrome session that had the inspect checkbox ticked.

## TL;DR

On a machine where the user's daily browser is a non-Chrome Chromium (Dia, Arc, Edge, ...) but Chrome is also installed, the harness (1) lets a stale Chrome `DevToolsActivePort` shadow the healthy target browser during discovery, and (2) on failure force-opens **Google Chrome** specifically. Combined effect in our case: the agent ended up silently automating a freshly-launched Chrome (without the user's logins) instead of the user's actual browser, and Chrome kept popping up on `chrome://inspect/#remote-debugging` during the task.

## What happened (real-world sequence)

1. Agent runs a task. Dia is healthy on 9222 with a valid `DevToolsActivePort` in its user-data dir.
2. `get_ws_url()` (daemon.py) walks `PROFILES` in order. Chrome's dir ranks **above** Dia's, so the stale Chrome file is matched first. The code then polls the dead port for 30s and **raises** ("DevTools is not live yet") — it never advances to Dia's entry later in `PROFILES`.
3. `ensure_daemon()` matches the error via `_needs_chrome_remote_debugging_prompt()` and calls `_open_chrome_inspect()`, which on macOS runs `tell application "Google Chrome" ...` — hardcoded. Wrong browser for Dia/Arc/Edge users; the user watches Chrome open on the inspect page repeatedly (once per cold harness invocation).
4. Because Chrome's inspect checkbox happened to still be ticked, the freshly launched Chrome started a debug server and wrote a **new, live** `DevToolsActivePort`. From that moment Chrome wins every discovery, and the retry connects to Chrome. The task "succeeded" — in the wrong browser, without the user's logins.

## Bug 1: discovery stops at the first `DevToolsActivePort`, even if its port is dead

In `get_ws_url()`, the `PROFILES` loop raises after the 30s poll on the first file found instead of moving on to the next profile. A leftover file from any higher-ranked browser permanently shadows a healthy lower-ranked one. (Chromium does not clean up `DevToolsActivePort` on browser quit, so stale files are the norm, not the exception.)

**Suggested fix:** quick-probe each candidate (is anything listening on that port?) and skip dead ones, preferring a live endpoint over list order; only fall into the 30s wait/raise when no candidate is live.

## Bug 2: `_open_chrome_inspect()` hardcodes Google Chrome on macOS

For users whose browser is Dia/Arc/Edge, the consent checkbox lives in *their* browser, so opening Google Chrome is both ineffective and harmful (see step 4 above: it can mint a fresh live debug endpoint in the wrong browser). Related but distinct from #290 (Windows: `webbrowser.open()` can't handle `chrome://`).

**Suggested fix:** open the inspect page in the browser owning the profile dir that discovery matched (or at minimum the system default browser), instead of hardcoded Chrome.

## Bug 3 (minor): `_is_local_chrome_mode()` ignores `BU_CDP_URL`

It only checks `BU_CDP_WS`. A user who explicitly pins an endpoint via `BU_CDP_URL` still gets the chrome://inspect popup flow when the connection fails (e.g., while the per-connection "Allow remote debugging" consent dialog is pending in the target browser and the WS handshake times out).

**Suggested fix:** treat an explicit `BU_CDP_URL` like `BU_CDP_WS` — a user pinning an endpoint doesn't want the Chrome UI recovery flow.

## Workaround we use now

A pre-flight script reads the target browser's `DevToolsActivePort` and writes `BU_CDP_WS=ws://127.0.0.1:9222<path>` into the repo `.env`. That bypasses discovery entirely and (because `BU_CDP_WS` is set) disables the Chrome recovery flow. Works reliably, including across browser restarts (script refreshes the UUID and restarts the daemon when it changes).

Happy to send a PR for any/all of the three fixes if you're open to it.

citadelgrad:
Closes #425 (macOS browser-selection recovery path).

## Problem

`_open_chrome_inspect()` hardcodes Google Chrome through AppleScript. If that launch fails, it calls Python's generic `webbrowser.open()` with a `chrome://inspect` URL. On macOS that fallback can dispatch the unsupported URL to Safari, even when Microsoft Edge is the running Chromium browser.

That recovery is unusable—Safari cannot provide the Chromium CDP inspector—and it surprises the user by opening the wrong browser.

## Fix

- Reuse the existing `_BROWSER_LAUNCH` registry rather than adding another browser list.
- On macOS, prefer a running supported non-Chrome Chromium browser.
- Use `edge://inspect/#remote-debugging` for Edge and `chrome://...` for the other supported browsers.
- Use macOS `open` rather than AppleScript, avoiding Automation/TCC prompts.
- Probe candidate app installations with `open -Ra`.
- Fail closed when no supported Chromium browser can be opened; never send `chrome://` to Safari through the generic default-browser handler.
- Preserve the existing generic `webbrowser` behavior off macOS.

This supersedes the current-main portion of #471, which identified the hardcoded-browser problem and added the key running-browser/Edge-scheme behavior. That PR is currently conflicting with `main`; this version also covers the newer Safari fallback and reuses the browser registry now present on `main`. Credit to @ahmadalzaro1 for the original approach in #471.

## Tests

Red against `upstream/main`:

- `test_open_chrome_inspect_uses_running_edge_on_macos` fails because Chrome is hardcoded.
- `test_open_chrome_inspect_does_not_fall_back_to_safari_on_macos` fails because `webbrowser.open()` is called.

Green on this branch:

- `.venv/bin/pytest -q tests/unit/test_admin.py -k open_chrome_inspect` — 3 passed
- `.venv/bin/pytest -q tests/unit/test_admin.py` — 41 passed
- `.venv/bin/pytest -q` — 116 passed
- `git diff --check` — passed

Tested on macOS 26.4.1 with browser-harness 0.1.8 / browser-use 0.13.7 and Microsoft Edge installed and running.

---
## Summary by cubic
Open remote debugging in the active Chromium browser on macOS and never in Safari. Previously we AppleScript-launched Google Chrome and fell back to the default browser; now we detect a running supported Chromium app by exact process name and launch it with open, or fail closed. Also switch `_chrome_running` to exact process-name matching to avoid false positives.

- Select from `_BROWSER_LAUNCH` on macOS; prefer a running supported Chromium app (de-prioritize Chrome when Edge is running), else any installed candidate.
- Use `open` to launch `edge://inspect/#remote-debugging` for Edge channels and `chrome://inspect/#remote-debugging` otherwise; remove AppleScript and its TCC prompts.
- Detect installations without revealing Finder windows: check /Applications and ~/Applications, then fall back to Spotlight via `mdfind`.
- On macOS, never call `webbrowser.open`; return False if no supported Chromium is available. Preserve `webbrowser` behavior on other platforms.
- Add Chrome Beta/Dev and Edge Beta/Dev/Canary support; expand tests for exact process matching, Edge/Brave preference, Safari fallback blocking, and `_mac_app_installed` fallbacks.

<sup>Written for commit 68a222dc2258deb623319a969199bb92c44f92a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/600?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/browser_harness/admin.py
+++ b/src/browser_harness/admin.py
@@ -814,8 +814,13 @@ def _chrome_running():
 _BROWSER_LAUNCH = (
     # (profile-dir fragment, macOS app name, POSIX commands, Windows `start` target)
     ("chrome canary", "Google Chrome Canary", ("google-chrome-canary",), "chrome"),
+    ("chrome beta", "Google Chrome Beta", ("google-chrome-beta",), "chrome"),
+    ("chrome dev", "Google Chrome Dev", ("google-chrome-unstable",), "chrome"),
     ("chromium", "Chromium", ("chromium", "chromium-browser"), "chromium"),
     ("chrome", "Google Chrome", ("google-chrome-stable", "google-chrome"), "chrome"),
+    ("edge beta", "Microsoft Edge Beta", ("microsoft-edge-beta",), "msedge"),
+    ("edge dev", "Microsoft Edge Dev", ("microsoft-edge-dev",), "msedge"),
+    ("edge canary", "Microsoft Edge Canary", ("microsoft-edge-canary",), "msedge"),
     ("edge", "Microsoft Edge", ("microsoft-edge", "microsoft-edge-stable"), "msedge"),
     ("brave", "Brave Browser", ("brave-browser", "brave"), "brave"),
     ("arc", "Arc", (), None),
@@ -903,13 +908,14 @@ def _open_chrome_inspect():
     import platform, shutil, subprocess, webbrowser
     if platform.system() == "Darwin":
         try:
-            running = subprocess.check_output(
+            process_output = subprocess.check_output(
                 ["ps", "-A", "-o", "comm="], text=True, errors="replace", timeout=5,
-            ).lower()
+            )
+            running = {Path(line.strip()).name.lower() for line in process_output.splitlines() if line.strip()}
         except Exception:
-            running = ""
+            running = set()
 
-        apps = [(mac_app, "edge" if frag == "edge" else "chrome") for frag, mac_app, _, _ in _BROWSER_LAUNCH]
+        apps = [(mac_app, "edge" if frag.startswith("edge") else "chrome") for frag, mac_app, _, _ in _BROWSER_LAUNCH]
         apps.sort(key=lambda spec: (spec[0].lower() not in running, spec[0] == "Google Chrome"))
         for app, scheme in apps:
             if not shutil.which("open"):
--- a/src/browser_harness/admin.py
+++ b/src/browser_harness/admin.py
@@ -900,21 +900,39 @@ def _launch_browser():
 
 def _open_chrome_inspect():
     """Open chrome://inspect/#remote-debugging so the user can tick the checkbox."""
-    import platform, subprocess, webbrowser
-    url = "chrome://inspect/#remote-debugging"
+    import platform, shutil, subprocess, webbrowser
     if platform.system() == "Darwin":
         try:
-            r = subprocess.run([
-                "osascript",
-                "-e", 'tell application "Google Chrome" to activate',
-                "-e", f'tell application "Google Chrome" to open location "{url}"',
-            ], timeout=5, check=False, capture_output=True)
-            if r.returncode == 0:
-                return True
+            running = subprocess.check_output(
+                ["ps", "-A", "-o", "comm="], text=True, errors="replace", timeout=5,
+            ).lower()
         except Exception:
-            pass
+            running = ""
+
+        apps = [(mac_app, "edge" if frag == "edge" else "chrome") for frag, mac_app, _, _ in _BROWSER_LAUNCH]
+        apps.sort(key=lambda spec: (spec[0].lower() not in running, spec[0] == "Google Chrome"))
+        for app, scheme in apps:
+            if not shutil.which("open"):
+                break
+            try:
+                installed = subprocess.run(
+                    ["open", "-Ra", app], timeout=5, check=False, capture_output=True,
+                )
+                if installed.returncode != 0:
+                    continue
+                opened = subprocess.run(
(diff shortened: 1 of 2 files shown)


citadelgrad:
## Follow-up review fixes

Addressed two valid edge cases found during an additional review:

- Process detection now compares exact executable basenames from `ps`, so short browser names such as `Arc` and `Dia` cannot match unrelated processes such as `searchpartyd`.
- The shared launch registry now includes Chrome Beta/Dev and Microsoft Edge Beta/Dev/Canary, matching the channels already recognized by profile discovery. Edge channels use the `edge://` inspector scheme.

Added regression tests for the substring false-positive and Edge Beta selection.

Verification after the follow-up commit:

- `pytest -q tests/unit/test_admin.py -k open_chrome_inspect` — 5 passed
- `pytest -q` — 118 passed
- `git diff --check` — passed

PR #605, opened 2026-08-14
feat: add daemon identity to tab markers

### LINKED ISSUE #604: feat: identify parallel browser-harness workers in tab titles
## Problem

When multiple browser-harness daemons share one visible Chrome profile, the current tab marker is only `🐴` at the front of the title. The title does not show which daemon is attached, so operators cannot distinguish parallel workers.

Each daemon already has a caller-supplied `BU_NAME`, but that identity is not surfaced in the tab label.

## Proposal

Append the daemon identity to the page title:

`Example Domain | 🐴 [research-7f3a]`

The marker should:

- use the actual `BU_NAME`;
- append to the page title;
- update across navigations;
- avoid inventing a shared `worker-` prefix;
- remain independent of foreground/background activation policy.

Related: #375 and #582. Existing focus-preservation PRs cover activation behavior; this issue focuses on identifying the attached worker.

yusoofsh:
Fixes #604

## Summary

- append the daemon's `BU_NAME` to the horse tab marker;
- keep the marker stable across page navigations;
- remove/update the harness marker when switching targets;
- document distinct names for parallel daemons.

## Scope

This change labels the currently attached worker tab. It does not introduce tab locks, profile management, or focus-preservation behavior.

## Verification

- `uv run --with pytest pytest -q` — 117 passed
- Python compilation passed
- `git diff --check` passed

---
## Summary by cubic
Adds daemon identity to controlled tab titles so parallel daemons are easy to tell apart. Previously we prefixed a horse emoji; now we append " | 🐴 [BU_NAME]", keep it across navigations, and remove/update it when switching targets.

- Marker logic is centralized in `_tab_marker` and used by `daemon.start()` and `helpers._mark_tab()/switch_tab()`; application is idempotent and strips any prior marker (including the legacy prefix) before applying the new one.
- Required: set distinct `BU_NAME` values when running multiple daemons; the harness does not add a shared prefix.
- Behavioral note: the suffix is appended to existing titles (or becomes the title if empty); update any tooling that parses `document.title` to tolerate the new suffix.

<sup>Written for commit 8b8e41b5f89b1db8106d718a5c43733cf4f81be4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/605?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


subaoyan16:
Hi @yusoofsh, this is a really useful change — the daemon identity in tab titles is exactly what we need to tell parallel workers apart. I noticed the review comments have been waiting for about two weeks. I'd like to help get this landed. I've prepared an implementation that addresses all four findings: the marker JS consolidated into one shared module (no drift between daemon.py and helpers.py), the daemon's set_session marker switched to the suffix form so the double-marker race disappears, endsWith-based idempotency that also covers empty-title pages, and tests that assert the exact marker string instead of fragments. If you're still working on this, I'd be happy to share the diff or review yours. If you're busy, would you be open to me submitting a follow-up PR that fixes #604 on top of your approach, with credit to yours? Thanks!

yusoofsh:
hl @subaoyan16, ill get my agent working on this. Thanks for the heads up!

subaoyan16:
Got it, thanks for the reply!

Issue #136, opened 2026-04-21
tap handler in daemon.py blocks ~2–4s per navigation on await Runtime.evaluate


merijnvanes:
Thanks for the harness. Hit a performance thing worth flagging.

**Finding:**
On a fresh Chromium with cache-busted navigations, a `goto → wait_for_load → js(extract)` loop runs at ~4.5s per iteration. Playwright on the same browser does it in ~0.9s. One tweak gets the harness to ~1.06s, so it isn't the CDP layer.

Cause is `daemon.py:152`, the mark_js eval in the `tap` handler. Dropping the `await`:

```python
elif method in ("Page.loadEventFired", "Page.domContentEventFired"):
    asyncio.create_task(self.cdp.send_raw(
        "Runtime.evaluate", {"expression": mark_js}, session_id=self.session
    ))
```

Two stalls compound. `cdp-use` dispatches events one at a time, so the await on `domContentEventFired` holds up `loadEventFired` in the queue. Chrome's V8 thread is busy, the eval doesn't return, the 2s `wait_for` ceiling hits. Then the same thing happens on `loadEventFired`, and the user's next `js(...)` queues behind that second eval on V8. Two ~2s waits, one in event dispatch, one in Chrome.

Fire-and-forget fixes both. The 🟢 title still lands, just a moment later. Median per iteration: 4.595s → 1.060s.

**Question:**
The `await` and 2s cap both look deliberate. What I'm less sure is intentional: the cap trips twice per navigation (DCL + load), and the second eval stacks in V8's queue ahead of the user's next `js(...)`, so the cost doubles. Did you mean to pay that twice, or was one trip the worst case you had in mind? Happy to open a PR if the double-trip wasn't the intent.

**Env:** 
macOS 14.6 / Apple Silicon, Chrome for Testing 145, Python 3.13, `browser-harness@main`, `cdp-use==1.4.5`.

mvanhorn:
Implemented in PR https://github.com/browser-use/browser-harness/pull/171. Kept the existing `except Exception: pass` behavior via a module-level `_silent(coro)` wrapper so stale-session failures still swallow quietly, now off the critical path.

PR #608, opened 2026-08-14
Stop the tab marker from reaching what the agent reads

misterbridge:
The 🐴 tab marker is written into `document.title` — the same channel the page writes to — and nothing removes it before an agent reads it back. So the marker, which describes the harness, reaches the agent as if it described the site.

## Why this is worse than noise

The marker is not consistently present. Whether a title carries it is decided by a race between the harness and the page:

| Situation | Result |
|---|---|
| Page whose title is settled at load (plain SSR) | harness writes last → **marked** |
| Page that sets its title client-side after load (SPA, React hydration) | page writes last → **clean forever** |
| Tab reached by `switch_tab()` once already settled | harness writes last → **marked, whatever the page** |

Same URL, two access paths, opposite results:

```
after direct navigation, settled          clean   'Peru travel guide | Example'
after switch_tab() back onto that tab     MARKED  '🐴 Peru travel guide | Example'
```

The marker says nothing about the page — it says how the agent got there. That is exactly what makes it read as data: it correlates with real technical properties (render mode, load speed) without being one.

## This has already misled agents, in this repo

Six `domain-skills` files mention the marker. Four document the trap correctly. Two took it for a property of the site and built on it:

- `agent-workspace/domain-skills/vercel/vercel.md:135` — *"Page title prefix is a status signal — 🟢 Vercel = all systems nominal"*
- `agent-workspace/domain-skills/atlas/overview.md:70` — *"The app sets a green-dot emoji prefix on titles… Useful for `wait_for` conditions"*

Both are pre-existing and untouched here. The same question was raised from outside the project in [NousResearch/hermes-agent#85430](https://github.com/NousResearch/hermes-agent/issues/85430): *"It can be mistaken for actual website content during QA, so a documented explanation — or a less content-like marker / separate metadata field — would help prevent false findings."*

## What this changes

**No marker at all when the browser is headless.** With no window there is no one to read it, so it can only pollute. Detected from `Browser.getVersion`'s user agent at connection time.

**Every title a helper hands back is stripped** — `page_info()`, `current_tab()`, `list_tabs()`, `connection_status()`, and the title the recorder traces into `recording-summary.json`. Only a leading marker-plus-space is removed; `"Horses 🐴 for sale"` and `"🐴Emoji-first"` survive untouched.

Marking moves from `switch_tab()` into the daemon, which is what knows whether the browser has a window. The fire-and-forget marking on load events is preserved unchanged — awaiting it costs ~4 s per navigation (#136).

`SKILL.md` gains the warning that was missing: the marker exists, helper-returned titles are stripped of it, and `js("document.title")` therefore disagrees with `page_info()["title"]` in a headed session. Three domain-skill files that quoted the old marked titles are corrected.

## The follow-up, and why it comes second

The marker still vanishes when a page rewrites its own title after load — most modern sites, since an SPA or a hydrating framework sets its title after the load event fires. That is a separate bug and the fix is ready: [misterbridge/browser-harness-fork#1](https://github.com/misterbridge/browser-harness-fork/pull/1), six commits on top of this branch, opened there rather than here so its diff shows those six alone. It moves here once this lands.

The order is not a convenience. Making the marker persistent *without* the stripping in this PR would turn an intermittent leak into a permanent one: today the marker often disappears on its own, so some reads come back clean by accident; re-applied on every title write, it would reach every single read.

## Verified

- `pytest tests -q` → 130 passed.
- Live, headless Chrome 151: 0 of 25 title reads carried the marker, against 15 of 24 before the change — and the raw in-page `document.title` is never marked, confirming the skip fires rather than the stripping merely hiding it.

---
## Summary by cubic
Stops the 🐴 tab marker from reaching agent-visible titles while keeping it visible to users in headed sessions. Previously the harness wrote the marker into document.title and helpers returned it; now headless sessions never mark, and helper/recorded titles strip a leading "🐴 " prefix.

- Detect headless via Browser.getVersion once at connect; if undetermined, treat as headless. `BH_TAB_MARKER=0/1` forces off/on.
- Strip a leading "🐴 " from titles returned by page_info(), current_tab(), list_tabs(), connection_status(), and from recorder context.
- Move marking to the daemon and unmark the previous tab on session switch; client-side helpers no longer write to document.title.
- Match blank-page and startup-placeholder filters on stripped titles to avoid reusing the placeholder.
- Add `tab_marker` module with shared marker/JS and unit tests; update SKILL.md and domain skills.

**Migration notes**
- Remove waits/assertions that depend on a title prefix (e.g., 🟢/🐴); helpers no longer return it.
- To read the literal page title in headed runs, use js("document.title"); it can differ from page_info()["title"].

<sup>Written for commit 6f5e184e2370961a8e58501b7d81b1e46bd88079. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -162,7 +162,7 @@ If you get stuck on a browser mechanic, check https://github.com/browser-use/bro
 - `chrome://inspect/#remote-debugging` must be enabled for local Chrome control.
 - Chrome may show an "Allow remote debugging?" popup; wait for the user to click Allow. Do not retry in a loop — Chrome pops a fresh dialog for every new connection, and the daemon's single held connection is what makes this a one-time click.
 - Omnibox popups are not real work tabs.
-- A headed browser gets a `🐴 ` prefix on the driven tab's title, so the user can see which tab is yours. `page_info()`, `current_tab()` and `list_tabs()` return the page's own title, without it; raw reads like `js("document.title")` still see it.
+- A headed browser gets a `🐴 ` prefix on the driven tab's title, so the user can see which tab is yours. `page_info()`, `current_tab()` and `list_tabs()` return the page's own title, without it; raw reads like `js("document.title")` still see it. Headless browsers are never marked; `BH_TAB_MARKER=0` or `=1` forces the choice when detection gets it wrong.
 - CDP target order is not Chrome's visible tab-strip order.
 - `BU_CDP_URL` is an HTTP DevTools endpoint; the daemon resolves it to WebSocket.
 - Ask before leaving cloud browsers running; stop them with `stop_remote_daemon(name)` or `PATCH /browsers/{id} {"action":"stop"}`.
--- a/SKILL.md
+++ b/SKILL.md
@@ -162,6 +162,7 @@ If you get stuck on a browser mechanic, check https://github.com/browser-use/bro
 - `chrome://inspect/#remote-debugging` must be enabled for local Chrome control.
 - Chrome may show an "Allow remote debugging?" popup; wait for the user to click Allow. Do not retry in a loop — Chrome pops a fresh dialog for every new connection, and the daemon's single held connection is what makes this a one-time click.
 - Omnibox popups are not real work tabs.
+- A headed browser gets a `🐴 ` prefix on the driven tab's title, so the user can see which tab is yours. `page_info()`, `current_tab()` and `list_tabs()` return the page's own title, without it; raw reads like `js("document.title")` still see it.
 - CDP target order is not Chrome's visible tab-strip order.
 - `BU_CDP_URL` is an HTTP DevTools endpoint; the daemon resolves it to WebSocket.
 - Ask before leaving cloud browsers running; stop them with `stop_remote_daemon(name)` or `PATCH /browsers/{id} {"action":"stop"}`.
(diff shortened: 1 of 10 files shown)


misterbridge:
Both points are addressed in the three commits just pushed. Thanks — the first one was a real hole.

**Fail closed** (`daemon.py:376`): agreed, and fixed. When `Browser.getVersion` raises, or answers without a user agent, the session is now treated as headless and does not mark. The marker is cosmetic; the titles it pollutes are not, so an undetermined mode costs a horse rather than a leak.

**On the user-agent token** (`daemon.py:378`): the factual claim does not hold on current Chrome. Measured today on Chrome 151.0.7922.138 / macOS, launched `--headless=new`, read both over HTTP `/json/version` and over CDP `Browser.getVersion`:

```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/151.0.0.0 Safari/537.36
```

End to end, the skip does fire: in a headless session the raw in-page `document.title` reads `T-42`, never marked at all rather than stripped on the way out.

That said, the underlying concern is fair — this rests on one string the project does not control. I looked for a better signal and there is none:

| Signal | Verdict |
|---|---|
| `Browser.getBrowserCommandLine` | `Command line not returned because --enable-automation not set` — the harness attaches to browsers it did not launch |
| `SystemInfo.getInfo` | browser-target only, errors here |
| `screen.width/height` | 1920×1080 headed vs 800×600 headless, but a headless run can be sized to anything |
| `navigator.webdriver`, `colorDepth`, `plugins.length`, `navigator.connection`, `userAgentData.brands` | identical in both modes |

So the detection stays, and `BH_TAB_MARKER=0` / `=1` now settles it explicitly when the caller knows better than the browser does — same shape as the existing `BH_DOMAIN_SKILLS` / `BH_RECORD` switches. Verified on a real headless browser: unset and `0` leave `document.title` untouched, `1` marks it while `page_info()` still returns the page's own title.

PR #644, opened 2026-08-23
Reuse one tab for agent browsing instead of opening one per page

subzeroflame:
## The problem

`new_tab` opens a window per call. An agent sweeping a search result set calls it
in a loop, so a twenty page sweep leaves twenty tabs behind. In our use the user
closed dozens by hand twice in two days.

Telling the agent not to do it does not hold. A prompt is a request, and the loop
is in the code.

## The change

`new_tab(url)` now navigates a tab this harness already owns, and only creates a
target when there is nothing to reuse. `force=True` is there for the cases that
genuinely need a second tab, side by side comparison or a popup flow.

A tab the user opened is never taken over: reuse happens only for a tab the
harness itself opened, or for the attached tab when it is blank, which is the
behaviour that was already there.

## Why ownership is a file

Two simpler options do not work, and both fail in a way that looks fine at first.

Module state cannot carry it. Each heredoc invocation is a fresh process, so the
record is gone before the next call.

The title marker cannot carry it either. `_mark_tab` writes the horse prefix, but
`goto_url` triggers a load and the page replaces `document.title`, taking the
marker with it. A marker-based version reuses exactly once and then opens a tab
per page again, which is worse than no fix because it looks like it works.

So ownership lives in `config_dir()/agent-tab`, which survives both.

## Tests

Five cases, one per branch: reuse the recorded tab, record a newly created tab,
leave a user tab alone, create again when the recorded tab was closed, honour
`force`. The third is the one that catches the marker-based version.

`pytest tests/unit` passes, 128 tests.

## Verified against a live Chrome

Five navigations in one process, then three more in separate processes, all land
on one target with the tab count unchanged.

```
tabs before: 18
  https://example.com                     tabs=19  tid=02E32278C5FD
  https://example.org                     tabs=19  tid=02E32278C5FD
  https://www.iana.org/domains/reserved   tabs=19  tid=02E32278C5FD
  https://example.com/index.html          tabs=19  tid=02E32278C5FD
  https://www.rfc-editor.org/rfc/rfc2606  tabs=19  tid=02E32278C5FD
distinct tabs used: 1
```

Before the change the same loop produced five tabs.

---
## Summary by cubic
Reuses a single agent-owned tab per daemon instead of opening a new tab on every call, reducing tab bloat and preventing cross-daemon interference. Previously each `new_tab` created a tab; now `new_tab(url)` navigates the owned tab and only creates when reuse is not possible.

- Behavior
  - For real URLs: reuse a live owned tab; otherwise reuse the attached tab if it is blank; otherwise create and record a new owned tab.
  - For `about:blank` or when `force=True`: always create a fresh tab.
  - Never takes over user tabs; only reuses a recorded owned tab or a blank attached tab.
  - Ownership is scoped per daemon under `config_dir()/agent-tab-{BU_NAME}` and is recreated if the recorded tab no longer exists.
  - Updates `interaction-skills/tabs.md` to instruct agents to loop on `new_tab(url)`, use `force=True` when two tabs are needed, and note that user tabs are never taken over.

- Concurrency and tests
  - Serializes select-or-create with a kernel advisory lock (flock on POSIX, `msvcrt.locking` on Windows); proceeds unlocked if not acquired within 5s; the OS releases the lock on process exit.
  - Tests cover reuse vs create, user-tab respect, per-daemon scoping, `force`, cross-process exclusivity, killed-holder release, proceed-unlocked fallback, and concurrent calls sharing one created tab. Subprocess tests inherit `PYTHONPATH`, use named child scripts to avoid `PYTHONHASHSEED` collisions, and launch the waiter on the holder’s “held” signal.

<sup>Written for commit 98f951506944ee9e39e1b71c06068bea23679200. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/644?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/interaction-skills/tabs.md
+++ b/interaction-skills/tabs.md
@@ -7,12 +7,19 @@ Use **CDP for control**, **UI automation for user-visible order**.
 ```python
 tabs = list_tabs()                    # includes chrome:// pages too
 real_tabs = list_tabs(include_chrome=False)
-tid = new_tab("https://example.com")  # create + attach in the background
+tid = new_tab("https://example.com")  # reuse this harness's tab, else create
+tid = new_tab("https://example.com", force=True)   # insist on a second tab
 switch_tab(tid)                       # attach harness, move the horse marker
 activate_tab(tid)                     # optional: explicitly show it in Chrome
 print(current_tab())
 print(page_info())
 ```
+
+Sweeping many pages is `new_tab(url)` in a loop, not a tab per page. It
+navigates the tab this harness already owns, so a twenty page sweep leaves one
+tab rather than twenty. `force=True` is for the cases that genuinely need two at
+once, such as comparing pages side by side or following a popup. A tab the user
+opened is never taken over.
 
 What CDP is good at:
 - attach to a tab
--- a/src/browser_harness/helpers.py
+++ b/src/browser_harness/helpers.py
@@ -294,44 +294,60 @@ def _owned_tab_path():
     return config_dir() / f"agent-tab-{ipc._check(NAME)}"
 
 
-def _owned_tab_lock(timeout=5.0):
+def _owned_tab_lock(wait=5.0, stale_after=120.0):
     """Serialise select-or-create so two agents do not each open a tab.
 
-    O_CREAT|O_EXCL rather than fcntl because the daemon runs on Windows too. A
-    lock older than the timeout is treated as abandoned, and a lock that cannot
-    be taken at all is proceeded past rather than raised: a crashed neighbour
-    should cost a duplicate tab at worst, never a failed call.
+    O_CREAT|O_EXCL rather than fcntl because the daemon runs on Windows too.
+
+    The two timeouts are deliberately different, and collapsing them into one
+    was a bug. The create path holds this across createTarget, attach and
+    goto_url, so a slow page load looks exactly like a crashed holder: a waiter
+    on the same threshold evicts a live holder, both run, and the duplicate tab
+    the lock exists to prevent comes back. `wait` is how long a caller queues,
+    `stale_after` is how old a lock must be before it is presumed abandoned, and
+    it sits well above any plausible hold.
+
+    Each holder stamps a token and removes only a lock still carrying it, so a
+    holder evicted despite that cannot delete its successor's lock on the way
+    out and let a third caller in.
     """
     import contextlib
+    import secrets
 
     @contextlib.contextmanager
     def _cm():
         path = _owned_tab_path().with_name(_owned_tab_path().name + ".lock")
-        deadline = time.time() + timeout
-        fd = None
+        token = f"{os.getpid()}-{secrets.token_hex(8)}"
+        deadline = time.time() + wait
+        held = False
         while True:
             try:
                 fd = os.open(str(path), os.O_CREAT | os.O_EXCL | os.O_WRONLY)
+                try:
+                    os.write(fd, token.encode())
+                finally:
+                    os.close(fd)
+                held = True
                 break
             except FileExistsError:
                 try:
-                    if time.time() - os.path.getmtime(path) > timeout:
+                    if time.time() - os.path.getmtime(path) > stale_after:
                         os.unlink(path)
-                        continue
+                        continue  # race for it again; the winner stamps its own
                 except OSError:
                     pass
                 if time.time() >= deadline:
-                    break
+                    break  # proceed unlocked rather than fail the call
                 time.sleep(0.05)
             except OSError:
                 break
         try:
             yield
         finally:
-            if fd is not None:
-                os.close(fd)
+            if held:
                 try:
-                    os.unlink(path)
+                    if path.read_text() == token:
+                        os.unlink(path)
                 except OSError:
                     pass
 
--- a/src/browser_harness/helpers.py
+++ b/src/browser_harness/helpers.py
@@ -288,7 +288,54 @@ def current_tab():
 
 def _owned_tab_path():
     from .paths import config_dir
-    return config_dir() / "agent-tab"
+    # Named daemons share one Chrome, so the record has to be per BU_NAME or one
+    # agent adopts another's live tab and navigates it out from under it. Same
+    # convention the recorder uses for its active marker.
+    return config_dir() / f"agent-tab-{ipc._check(NAME)}"
+
+
+def _owned_tab_lock(timeout=5.0):
+    """Serialise select-or-create so two agents do not each open a tab.
+
+    O_CREAT|O_EXCL rather than fcntl because the daemon runs on Windows too. A
+    lock older than the timeout is treated as abandoned, and a lock that cannot
+    be taken at all is proceeded past rather than raised: a crashed neighbour
+    should cost a duplicate tab at worst, never a failed call.
+    """
+    import contextlib
+
+    @contextlib.contextmanager
+    def _cm():
+        path = _owned_tab_path().with_name(_owned_tab_path().name + ".lock")
+        deadline = time.time() + timeout
+        fd = None
+        while True:
+            try:
+                fd = os.open(str(path), os.O_CREAT | os.O_EXCL | os.O_WRONLY)
+                break
+            except FileExistsError:
+                try:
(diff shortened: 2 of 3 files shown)


subzeroflame:
Both valid, both fixed in 7c106f1.

**P1, ownership scoped by `BU_NAME`.** This was the one that mattered. Named daemons share a Chrome, so a single record let one agent adopt another's live tab and navigate it away mid-task, which is worse than the tab pile-up this PR set out to fix. The record is now `agent-tab-{BU_NAME}`, matching the convention `recorder.py` already uses for its active marker, and it goes through the existing `_ipc._check` traversal guard so a hostile name cannot walk out of the config dir.

**P2, the create path is serialised.** Correct that the file write alone did not close it: two processes could each find nothing to reuse and each open a tab. Select-and-create now runs under a lock. Three notes on the shape:

- `O_CREAT|O_EXCL` rather than `fcntl`, since the daemon runs on Windows too.
- A lock older than the timeout is treated as abandoned, so a crashed process does not cost the next one its reuse.
- If the lock cannot be taken at all, the call proceeds unlocked rather than raising. A duplicate tab is a much smaller failure than a browsing call that throws.

Tests added, one per claim: the record differs across two `BU_NAME` values, a traversing name raises, a stale lock is broken rather than waited out, and four concurrent `new_tab` calls create exactly one target. `pytest tests/unit` passes at 128.

Also re-verified live: three sequential navigations still land on one tab, and `BU_NAME=alpha` and `BU_NAME=beta` get separate records.

subzeroflame:
Valid, and you were right that it recreated the bug the lock exists to prevent. Fixed in e76531d.

The root cause was one parameter doing two jobs. `timeout` was both the wait budget and the staleness threshold, and the create path holds the lock across `Target.createTarget`, attach and `goto_url`, so a slow page load was indistinguishable from a crashed holder. Your interleaving follows from that: waiter evicts a live holder, holder's `finally` then unlinks what is by then the successor's lock, third caller walks in.

Two changes:

- The timeouts are separate. `wait=5.0` is how long a caller queues; `stale_after=120.0` is how old a lock must be before it is presumed abandoned, comfortably past any plausible hold rather than sitting on top of it.
- A holder stamps `pid-randomhex` into the lock on acquisition and removes it only while it still carries that token, so a holder evicted anyway cannot take its successor's lock with it.

Three tests, one per branch of the fix: a genuinely stale lock is still broken without waiting, a slow live holder is left alone and the waiter proceeds unlocked instead of evicting it, and a holder whose lock was taken over exits without deleting it. `pytest tests/unit` passes at 134, and three live navigations still land on one tab.

One thing I left deliberate: if the lock cannot be acquired within `wait`, the call proceeds without it rather than raising. The worst case is a duplicate tab, which is a much smaller failure than a browsing call that throws, and it keeps a wedged neighbour from taking the harness down with it.

subzeroflame:
Valid, and it is the same window from the other side. Fixed in 2bbe97a by taking your second option: locks whose holder can still exit are never evicted.

You are right that there is no atomic compare-and-unlink to close the read-then-unlink gap with, so I stopped trying to close it and removed the thing that opens it. Age no longer reclaims anything. A lock is taken only when the process that stamped it is gone, which a living holder cannot be, so a holder always finds its own stamp on the way out and the successor case cannot arise.

Liveness is `os.kill(pid, 0)` on the pid stamped into the lock, and every uncertain answer is a yes: a pid owned by another user raises `PermissionError` and the lock stays, and a platform without `os.kill` never reclaims at all. In both cases callers wait and then run unlocked, so the worst case stays the duplicate tab this lock exists to avoid, never a wedged caller.

This also drops `stale_after`, which I was glad to lose. It required guessing a plausible hold time for a section that spans `goto_url`, and that guess was the root of the previous two rounds.

Five tests: a dead holder's lock is reclaimed and released, an ancient but live holder's is left alone, a holder whose lock was taken over anyway exits without deleting it, and both liveness answers. `pytest tests/unit` passes at 136. Live: three navigations land on one tab with no lock left behind, and a planted dead-holder lock is reclaimed and released.

Thanks for staying on this one. All three rounds were real, and the first was the one that would have hurt users.

subzeroflame:
Both valid, and both were mine to own. Fixed in 2fc7e6a by taking your other option: the lock is an OS advisory lock now, not a file I interpret.

The two findings share a cause. A crash between `os.open` and `os.write` leaves the record unwritten, and a reused pid makes a dead holder look alive, because ownership was something I encoded into a file's contents and then had to reason about. Every round of this has been one more hole in that reasoning: a shared timeout evicting a live holder mid `goto_url`, then a release racing a successor with no atomic compare-and-unlink available, and now these two.

So the lock is `flock` on POSIX and `msvcrt.locking` on Windows, held on an open descriptor. The kernel releases it when the holder exits, however it exits, which answers staleness, pid identity and the unwritten record together instead of one per round. The file is only a handle to lock. Nothing unlinks it, so no caller can delete a lock another one is holding, and a leftover or empty file means nothing to anybody.

One deliberate choice remains, and I want it on the record rather than found later: a lock that cannot be taken within `wait` is run past unlocked rather than raised on. The worst case is the duplicate tab this exists to avoid, and a wedged neighbour should not take the harness down with it. Happy to make it raise instead if you would rather it be strict.

Four tests, two using real subprocesses since the interesting properties are cross-process: a second process waits rather than running alongside the first, a killed holder does not block the next caller, an unavailable lock proceeds instead of raising, and the file is never unlinked. `pytest tests/unit` passes at 135. Live: three navigations in one process and two more across separate processes all land on one tab.

subzeroflame:
Correct, and it was mine. This is a src layout, so the child scripts pointed at the repo root when the package lives one level down. They passed here only because an editable install happened to have `src` on the path; on a fresh checkout they would have failed with `ModuleNotFoundError`. I reproduced it both ways under an isolated interpreter on a copy of the tree with nothing installed:

```
old parents[2]     -> rc=1 ModuleNotFoundError: No module named 'browser_harness'
src on path        -> rc=0 imported
```

Fixed in a0f2bcc, though not by moving the index. The children now inherit the parent's `sys.path` through `PYTHONPATH`, so nothing in a test computes where the package lives and the layout can move without the tests noticing. They also carry timeouts and surface the child's stderr, since the previous shape could hang if a kill were ever missed and told you nothing when it broke.

Verified the fix against the exact condition you named: `pytest tests/unit/test_helpers.py -k owned_tab_lock` passes under `uv run --isolated` on a copy of the tree with no install.

Two tidies in the same area while I was there: the lock helper resolves the record path once rather than twice, and the platform lock primitive is chosen at import rather than re-imported on every acquire, which also makes it visible which implementation a build is on.

135 unit tests pass. Live: three navigations still land on one tab.

subzeroflame:
Valid, and worse than it looks at first. Because `hash()` is PYTHONHASHSEED-randomised, the collision is not simply possible, it is possible on some runs and not others, and when it lands both children run the same body and the exclusivity assertion passes for the wrong reason. A test that can quietly stop testing anything is worse than one that fails. Fixed by having the caller name them.

I also fixed the other half of the same test, which you did not flag but has the same shape. It launched the waiter after a 0.3 second sleep, so on a loaded machine the waiter could take the lock first and the assertion would fail for a reason unrelated to the code. The holder now announces that it is inside before it starts holding, and the waiter is launched on that line rather than on a timer.

135 unit tests pass. The lock tests pass five consecutive runs with a randomised hash seed, and once more under `uv run --isolated` on a copy of the tree with nothing installed.

subzeroflame:
Fair, and worth fixing rather than waving off. The print sits inside the locked block, so the line means the lock is already taken, and my wording described it as happening before. Read the old way, someone could move the print above the `with` and the waiter would then be racing intent rather than possession, which is the same vacuous assertion the previous commit removed.

Reworded in the latest commit, with the reason to keep the print where it is stated in the comment rather than left implicit.

135 unit tests pass.

PR #655, opened 2026-08-26
feat: add pt-BR AI dubbing to video export

thetunnes:
Adds `dubbing.py`: collapses each beat's sticky narration text into timed spans, synthesizes each span via the ElevenLabs TTS API (eleven_multilingual_v2, pt), and mixes them into a single WAV sized to the compiled video duration using ffmpeg's adelay/amix filters.

`video export` gains a `--dub` flag that mixes this narration track into the exported MP4 (aac) instead of leaving it silent. Requires ELEVENLABS_API_KEY (see .env.example); ELEVENLABS_VOICE_ID optionally overrides the stock voice. Only action-beat narration is voiced, not the intro/outcome cards.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,7 @@
 # Copy to .env — auto-loaded. Only needed for remote browsers.
 BROWSER_USE_API_KEY=bu_your_key_here
+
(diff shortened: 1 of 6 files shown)


MagMueller:
who are you making the narrated browser videos for?

PR #674, opened 2026-08-29
domain-skills/linkedin: exact company headcount from the Insights tab

Altimor:
Adds `domain-skills/linkedin/company-insights.md`. Field-tested 2026-08-28 on a free (non-Premium) account.

The headline finding is a silent-wrong-data trap. The exact headcount on a company's Insights tab **lazy-hydrates**: read it as soon as the tab opens and `/company/langchain/` reports **133**; scroll the page first and the same DOM node settles at **423**. Nothing throws and nothing looks stale, so the wrong number gets recorded as fact. `wait_for_load()` does not cover it — the document completes long before the figure settles.

Also documented:

- `/company/<slug>/insights/` **silently redirects** to the company home, so the tab has to be clicked. The URL rewrites with no error, so a naive `goto` looks like it worked and then scrapes the overview page.
- The click itself races the nav bar becoming interactive — the anchor is clickable before its handler is wired, so wait for the URL to change rather than a fixed timeout.
- Anchor extraction on `Total employee count`, not `Functional distribution` (the latter is the node that shows the placeholder), and note the value sits *before* the literal `total employees` header row.
- A free tripwire: range-check the result against the `201-500 employees` bucket in the header, which is server-rendered and never hydrates late. `133` falls outside `201-500`, so the bad read raises instead of passing silently.
- Profiles frequently render **no Experience section** (server-driven layout), which makes `a[href*="/company/"]` actively misleading — on one profile it returns `programmier-bar`, a podcast, rather than the actual employer.

No secrets, cookies, session tokens, or user-specific state; no pixel coordinates.

---
## Summary by cubic
Adds a `domain-skills/linkedin/company-insights` guide for extracting exact company headcount from the Insights tab on a free account. The figure lazy-hydrates and silently returns a stale placeholder (e.g., 133 for LangChain) unless the page is scrolled first (correct value 423).

- Scrolls the page before extraction and range-checks the result against the server-rendered header bucket (e.g., 201-500 employees) to catch premature reads.
- Navigates directly to `/company/<slug>/insights/`; clicking the tab is fragile because the anchor is visible before its handler is wired.
- Anchors on `Total employee count` (not `Functional distribution`) and warns that profiles often omit the Experience section, making `a[href*="/company/"]` unreliable for finding the employer.

<sup>Written for commit 3985326b1ffc5763b046dae400f18098764692ab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/674?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/domain-skills/linkedin/company-insights.md
+++ b/domain-skills/linkedin/company-insights.md
@@ -1,0 +1,116 @@
+# LinkedIn — exact company headcount from the Insights tab
+
+Field-tested against linkedin.com on 2026-08-28, signed in with a free (non-Premium) account.
+
+The company **header** only ever shows a self-reported bucket — `201-500 employees`. The **exact**
+headcount lives on the company's Insights tab. It is not Premium-walled on a free account.
+
+## The trap: the number lazy-hydrates, and the stale value looks real
+
+This is the whole reason this file exists. Read the Insights tab as soon as it opens and you get a
+number that is **wrong, plausible, and silent** — nothing throws, nothing looks stale.
+
+For `/company/langchain/`:
+
+| when you read | "Functional distribution" says |
+|---|---|
+| immediately after the tab opens | `133` |
+| after scrolling to the bottom | `423`  ← correct |
+
+Same DOM node, same selector. The first render is a partial-data placeholder that is replaced once
+the lazy sections below the fold hydrate. `wait_for_load()` does **not** cover this — the document is
+complete long before the figure settles.
+
+**Always scroll the whole page before extracting**, then read the value under the `Total employee
+count` heading (see below), not the first number you can match.
+
+## Route: the Insights tab must be clicked, not navigated to
+
+`https://www.linkedin.com/company/<slug>/insights/` **silently redirects back to the company home.**
+There is no error and the URL rewrites, so a naive `goto` looks like it worked and then scrapes the
+overview page instead.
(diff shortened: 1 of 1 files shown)


Altimor:
Correction pushed — I had one claim backwards in the original version.

I wrote that `/company/<slug>/insights/` silently redirects to the company home and that the Insights tab therefore has to be clicked. **That does not reproduce.** Direct navigation lands correctly on every company I tried, with and without the trailing slash. The original observation was made while my cookie export was stale, which I think explains it.

The advice was in fact inverted: **clicking is the fragile path.** The nav anchor is in the DOM and reports `visible` before its handler is wired, so the click lands, does nothing, and leaves you scraping the overview page — URL unchanged, no error raised. Three consecutive clicks on a present, visible anchor failed exactly that way on `/company/mem0/`, which is what prompted the retest.

The lazy-hydration finding (133 → 423 for LangChain) is unaffected and still the main point of the file.

PR #701, opened 2026-08-29
fix: decode http_get() with the encoding the response declares

### LINKED ISSUE #683: `http_get()` hard-codes UTF-8 and ignores the response charset — `UnicodeDecodeError` on any non-UTF-8 page
**Labels:** bug, helpers

## Description

`http_get()` is the documented "no browser needed" path — `SKILL.md`'s *When Not to Use*
section tells the agent to prefer a plain fetch for public pages. It ends with an
unqualified `.decode()`:

```python
# src/browser_harness/helpers.py:535-541
import gzip
h = {"User-Agent": "Mozilla/5.0", "Accept-Encoding": "gzip"}
if headers: h.update(headers)
with urllib.request.urlopen(urllib.request.Request(url, headers=h), timeout=timeout) as r:
    data = r.read()
    if r.headers.get("Content-Encoding") == "gzip": data = gzip.decompress(data)
    return data.decode()
```

`bytes.decode()` with no argument is `utf-8`, strict. The response's own
`Content-Type: text/html; charset=...` is available on `r.headers` and is never consulted.

## Impact

Two failure modes, both on perfectly ordinary sites:

1. **Hard crash** on any page served as latin-1 / windows-1252 / gbk / shift_jis /
   euc-kr — still common on regional news, government, and older corporate sites:

   ```
   UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 1234: invalid continuation byte
   ```

   The exception propagates straight out of the user's script. There is no fallback and
   no error message pointing at the charset, so the agent's most likely next move is to
   spin up a whole browser for a page a `.decode(charset)` would have read.

2. **Silent mojibake** where the bytes happen to be valid UTF-8 but the document is
   declared as something else.

Note the module already carries scar tissue from a related bug — `run.py:4-11` forces
UTF-8 on stdout/stderr for exactly this class of problem — so the codebase has clearly
been bitten here before.

## Reproduction

```python
# any windows-1252 page, e.g. a legacy site with an é in the body
print(http_get("http://example-legacy-site.tld/"))
```

## Suggested fix

`urllib`'s response already parses the header. One line:

```python
return data.decode(r.headers.get_content_charset() or "utf-8", errors="replace")
```

`get_content_charset()` returns the `charset=` parameter from `Content-Type` or `None`.
`errors="replace"` keeps a mislabelled page from crashing the run — a page with a few
replacement characters is far more useful to an agent than a traceback.

Worth noting the `fetch_use` branch above it returns `.text`, which does handle charset,
so today `http_get()` behaves differently depending on whether `BROWSER_USE_API_KEY`
happens to be set. That inconsistency goes away with the same fix.

rajarshidattapy:
Fixes #683.

## Problem

`http_get()` ended in a bare `data.decode()` — UTF-8, strict. The response's own `Content-Type: ...; charset=` was sitting on `r.headers` and was never consulted:

```python
data = r.read()
if r.headers.get("Content-Encoding") == "gzip": data = gzip.decompress(data)
return data.decode()
```

So any page served as windows-1252, gbk, shift_jis or euc-kr — still common on regional news, government and older corporate sites — raised straight out of the caller's script:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 3
```

No fallback, and nothing in the error naming the charset. `SKILL.md` points at this helper specifically so agents *don't* launch a browser for a page a plain fetch can read, so the likely recovery from an unexplained traceback is the browser it exists to avoid.

## Fix

Resolve the encoding the way the HTML spec does, in `_response_encoding()`:

1. **BOM** — wins outright (`utf-8-sig` strips it rather than leaving a stray `\ufeff` that breaks JSON parsing and string comparisons).
2. **`Content-Type` charset** — via `headers.get_content_charset()`.
3. **`<meta charset>` prescan** of the first 2 KB, covering both spellings the spec allows.
4. **UTF-8** otherwise.

A charset no codec matches falls back to UTF-8 instead of raising, and the decode uses `errors="replace"` so a mislabelled page yields a few U+FFFD rather than killing the fetch.

**On the meta step** — the issue suggested a header-only fix, and I went past it deliberately. A legacy page that declares `gbk` in markup and says nothing in its headers is the common real-world shape, and header-only would have turned that crash into silent mojibake rather than fixing it. That is the second failure mode the issue lists, so a one-line header fix would only have covered half of it.

This also settles an inconsistency the issue notes: the `fetch_use` branch above returns `.text`, which is already charset-aware, so `http_get()` previously behaved differently depending on whether `BROWSER_USE_API_KEY` happened to be set.

## Verification

Seven new tests in `test_helpers.py`. Five fail against the unfixed helper, reproducing the reported error exactly; the other two are the already-correct paths, guarding the new code against regressing them:

```
$ git stash -- src/browser_harness/helpers.py && pytest tests/unit/test_helpers.py -k http_get -q
E  UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 3: unexpected end of data
FAILED  test_http_get_honours_the_content_type_charset
FAILED  test_http_get_falls_back_to_the_meta_charset
FAILED  test_http_get_strips_a_utf8_bom
FAILED  test_http_get_replaces_undecodable_bytes_instead_of_raising
FAILED  test_http_get_decodes_after_gunzip
5 failed, 2 passed

$ git stash pop && pytest tests/unit/test_helpers.py -k http_get -q
7 passed
```

Checked the resolution table directly, including the false-positive case that worried me — a `charset=` appearing inside an unrelated `content` attribute:

```
utf-8       <- <meta charset="utf-8">
utf-8       <- <meta charset=utf-8>
gbk         <- <meta http-equiv="Content-Type" content="text/html; charset=gbk">
Shift_JIS   <- <META CHARSET='Shift_JIS'>
euc-kr      <- <meta charset="euc-kr">
utf-8       <- <meta name="description" content="about charset=bogus stuff">   # rejected by codecs.lookup
utf-8       <- <html>no meta</html>
utf-8       <- <meta charset="gbk">  with header charset=utf-8                 # header wins
```

Full suite: `9 failed, 159 passed` → `9 failed, 166 passed`. The 9 are pre-existing and unrelated — 8 POSIX-only scaffolding in `test_admin.py` (#680) and `test_skill.py` (#678, fixed in #693).

## Notes for reviewers

- The meta prescan is 2 KB rather than the spec's 1 KB, which is more generous and costs nothing since the body is already in memory.
- A bogus `charset=` inside an unrelated attribute can only ever match when there is no header charset, and any value that isn't a real codec falls back to UTF-8 — so the worst case is the behaviour you have today.
- `_response_encoding` runs *after* gunzip, so it prescans the real document rather than compressed bytes. `test_http_get_decodes_after_gunzip` pins that ordering.
- The `fetch_use` branch is untouched.

---
## Summary by cubic
Fixes #683 by making `http_get()` decode responses using the encoding the response declares — BOM (UTF-8/16/32), `Content-Type` charset, `<meta charset>` prescan, then UTF-8 — instead of always assuming strict UTF-8. Also decodes with `errors="replace"` so mislabelled pages return replacement characters rather than raising.

- Handles legacy pages that declare charset only in `<meta>` (the common case) and falls back to UTF-8 for unknown charsets.
- Adds tests covering each resolution step, the fallback, replace behaviour, gunzip decoding order, and all UTF BOMs — including a pin on the UTF-32LE-before-UTF-16LE branch order that prevents silent corruption.

<sup>Written for commit 0361d33a445c39be30e7ffe53d6610c9df4af872. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/701?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/browser_harness/helpers.py
+++ b/src/browser_harness/helpers.py
@@ -3,7 +3,7 @@
 Core helpers live here. Agent-editable helpers live in
 BH_AGENT_WORKSPACE/agent_helpers.py.
 """
-import base64, importlib.util, json, math, os, sys, time, urllib.request
+import base64, codecs, importlib.util, json, math, os, re, sys, time, urllib.request
 from pathlib import Path
 from urllib.parse import urlparse
 
@@ -521,8 +521,48 @@ def upload_file(selector, path):
     if not nid: raise RuntimeError(f"no element for {selector}")
     cdp("DOM.setFileInputFiles", files=[path] if isinstance(path, str) else list(path), nodeId=nid)
 
+# Charset from an HTML <meta> tag, in either of the two spellings the HTML spec
+# allows. Only the head of the document is scanned, which is where both must appear.
+_META_CHARSET = re.compile(
+    rb"""<meta[^>]+?charset\s*=\s*["']?\s*([a-zA-Z0-9_\-:.]+)""",
+    re.IGNORECASE,
+)
+_META_PRESCAN_BYTES = 2048
+
+
+def _response_encoding(data, response_headers):
+    """Encoding a response's bytes are actually in.
+
+    Resolution order follows the HTML spec: a BOM wins outright, then the
+    Content-Type charset, then a <meta charset> in the head, then UTF-8. Skipping
+    the last two would leave the common case — a legacy page that declares gbk or
+    shift_jis in markup and says nothing in its headers — silently mojibaked.
+    """
+    if data.startswith(codecs.BOM_UTF8):
+        return "utf-8-sig"
+    if data.startswith((codecs.BOM_UTF16_LE, codecs.BOM_UTF16_BE)):
+        return "utf-16"
+    declared = response_headers.get_content_charset() if response_headers else None
+    if not declared:
+        match = _META_CHARSET.search(data[:_META_PRESCAN_BYTES])
+        if match:
+            declared = match.group(1).decode("ascii", errors="replace")
+    if not declared:
+        return "utf-8"
+    try:  # a bogus or misspelled charset must not take the whole fetch down
+        codecs.lookup(declared)
+    except LookupError:
+        return "utf-8"
+    return declared
+
+
 def http_get(url, headers=None, timeout=20.0):
     """Pure HTTP — no browser. Use for static pages / APIs. Wrap in ThreadPoolExecutor for bulk.
+
+    Decoded using the encoding the response declares (BOM, Content-Type charset,
+    then <meta charset>), falling back to UTF-8. Undecodable bytes become U+FFFD
+    rather than raising — a page with a few replacement characters is more useful
+    than a traceback.
 
     When BROWSER_USE_API_KEY is set, routes through the fetch-use proxy (handles bot
     detection, residential proxies, retries). Falls back to local urllib otherwise."""
@@ -538,7 +578,7 @@ def http_get(url, headers=None, timeout=20.0):
     with urllib.request.urlopen(urllib.request.Request(url, headers=h), timeout=timeout) as r:
         data = r.read()
         if r.headers.get("Content-Encoding") == "gzip": data = gzip.decompress(data)
-        return data.decode()
+        return data.decode(_response_encoding(data, r.headers), errors="replace")
 
 
 # Imported at the bottom so recorder's own `from . import helpers` sees a
--- a/src/browser_harness/helpers.py
+++ b/src/browser_harness/helpers.py
@@ -543,7 +543,12 @@ def _response_encoding(data, response_headers):
     if data.startswith((codecs.BOM_UTF16_LE, codecs.BOM_UTF16_BE)):
         return "utf-16"
     declared = response_headers.get_content_charset() if response_headers else None
-    if not declared:
+    media_type = response_headers.get_content_type() if response_headers else None
+    if not declared and (
+        not response_headers
+        or response_headers.get("Content-Type") is None
+        or media_type in ("text/html", "application/xhtml+xml")
+    ):
         match = _META_CHARSET.search(data[:_META_PRESCAN_BYTES])
         if match:
             declared = match.group(1).decode("ascii", errors="replace")
--- a/src/browser_harness/helpers.py
+++ b/src/browser_harness/helpers.py
@@ -540,6 +540,8 @@ def _response_encoding(data, response_headers):
     """
     if data.startswith(codecs.BOM_UTF8):
         return "utf-8-sig"
+    if data.startswith((codecs.BOM_UTF32_LE, codecs.BOM_UTF32_BE)):
+        return "utf-32"
     if data.startswith((codecs.BOM_UTF16_LE, codecs.BOM_UTF16_BE)):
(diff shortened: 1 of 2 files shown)


rajarshidattapy:
Valid on both counts — the coverage gap and the ordering hazard. Confirmed the collision is exactly as described:

```
BOM_UTF32_LE b'\xff\xfe\x00\x00'
BOM_UTF16_LE b'\xff\xfe'
UTF32_LE startswith UTF16_LE BOM -> True
UTF32_BE startswith UTF16_BE BOM -> False
```

So it's a live hazard for little-endian only. A UTF-32LE body run through the UTF-16 branch decodes without raising and returns NUL-interleaved garbage:

```
utf-32 doc -> decoded as utf-16 -> '\x00h\x00é\x00l\x00l\x00o\x00'
```

Silent corruption, no exception, which is the worst shape for this. Added:

- a parametrised round trip through `http_get()` for all four BOMs (UTF-16 LE/BE, UTF-32 LE/BE);
- `test_utf32_bom_is_checked_before_utf16`, which pins the branch order, asserts both prefix relationships so the LE-only nature is explicit, and spells out what the reordered version would have produced;
- the reverse check — a UTF-16LE body must not trip the UTF-32 branch (it doesn't; `\xff\xfe` followed by a real character never matches `\xff\xfe\x00\x00`).

Verified the guard actually bites by swapping the two branches in `_response_encoding()`:

```
FAILED test_http_get_decodes_every_utf_bom[utf-32-le]
    assert '\x00h\x00é\x00l...' == 'héllo 中文'
FAILED test_utf32_bom_is_checked_before_utf16
    assert 'utf-16' == 'utf-32'
```

The other three BOM cases keep passing under the swap, which is the evidence that LE is the only ordering that matters.

No source change in this commit — the order you pushed is correct, it just wasn't pinned. Suite is `9 failed, 172 passed`, the 9 being the pre-existing unrelated ones (#680, #678).

Two notes while I was in here, both about the web edits on this branch rather than your comment:

- Restricting the `<meta charset>` prescan to `text/html` / `application/xhtml+xml` is a good tightening, and the `response_headers.get("Content-Type") is None` guard correctly handles `email.message.Message` defaulting `get_content_type()` to `text/plain` when the header is absent. My existing meta test covers the HTML path; the negative case (a `text/plain` response with a `charset=` in the body being ignored) is still uncovered if you want it pinned too.
- Worth confirming on a POSIX checkout — everything here was verified on Windows.

Pushed as `0361d33`.

MagMueller:
found a regression on 0361d33: a UTF-8 page with <meta name="description" content="about charset=windows-1252"> changes café to cafÃ©. reproduced through real local HTTP. can we parse actual meta attributes instead of matching arbitrary charset text?

PR #702, opened 2026-08-29
fix: don't assume uv when updating a wheel install

### LINKED ISSUE #688: `browser-harness --update` crashes with `FileNotFoundError` when `uv` isn't installed
**Labels:** bug, cli, ux

## Description

`run_update()` branches on install mode. The `pypi` branch shells out to `uv`
unconditionally:

```python
# src/browser_harness/admin.py:1229-1232
elif mode == "pypi":
    tool_upgrade = subprocess.run(["uv", "tool", "upgrade", "browser-harness"])
    if tool_upgrade.returncode != 0:
        return tool_upgrade.returncode
```

There is no `shutil.which("uv")` guard — grepping the whole package, `"uv"` appears
exactly once, on that line. `_install_mode()` returns `"pypi"` for any non-git install
that reports a version (`admin.py:820-824`), which includes every `pip install
browser-harness` and `pipx install browser-harness`.

Neighbouring code in the same module does guard its external tools:

```python
# admin.py:1029, 1054
if not shutil.which("profile-use"):
    raise RuntimeError("profile-use not installed -- curl -fsSL https://browser-use.com/profile.sh | sh")
```

## Impact

A pip-installed user runs the command the tool itself told them to run. `run.py`'s help
advertises it, and `print_update_banner()` nags on it once a day:

```
[browser-harness] update available: 0.1.9 -> 0.1.10
[browser-harness] agents: run `browser-harness --update -y` to upgrade and restart the daemon
```

and gets a raw traceback:

```
Traceback (most recent call last):
  File ".../browser_harness/admin.py", line 1230, in run_update
    tool_upgrade = subprocess.run(["uv", "tool", "upgrade", "browser-harness"])
  ...
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

Worse for the advertised agent path: `--update -y` is documented as the *agent* command
(`browser-harness --update [-y]  pull the latest version (agents: pass -y)`). An agent
that runs it gets an unhandled exception, no diagnosis, and no idea that the fix is
"install a different package manager".

Note `pyproject.toml` does not require uv, `install.md` documents both uv and pip
installs, and nothing else in the runtime depends on uv — so this is the only place the
tool assumes a uv install.

## Reproduction

On a machine without `uv`:

```bash
pip install browser-harness
browser-harness --update -y
```

## Suggested fix

Detect the installer and fall back, rather than assuming:

```python
elif mode == "pypi":
    if shutil.which("uv"):
        cmd = ["uv", "tool", "upgrade", "browser-harness"]
    else:
        cmd = [sys.executable, "-m", "pip", "install", "--upgrade", "browser-harness"]
    r = subprocess.run(cmd)
    if r.returncode != 0:
        return r.returncode
```

If auto-upgrading a pip install is deliberately out of scope, then say so instead of
crashing — print the exact `pip install --upgrade browser-harness` line and return
non-zero. Either is fine; an unhandled `FileNotFoundError` is not.

rajarshidattapy:
Fixes #688.

## Problem

`run_update()`'s `pypi` branch shelled out to `uv` unconditionally:

```python
elif mode == "pypi":
    tool_upgrade = subprocess.run(["uv", "tool", "upgrade", "browser-harness"])
```

No `shutil.which()` guard — grepping the package, `"uv"` appears on exactly that one line, while the neighbouring `profile-use` calls in the same module *do* check before shelling out. And `_install_mode()` returns `"pypi"` for any non-git install that reports a version, which covers every `pip install browser-harness` and `pipx install browser-harness`.

So a user without uv, running the command the tool told them to run, got a traceback. `print_update_banner()` nags daily with `browser-harness --update -y`, and `HELP` advertises that same flag as *the agent path* — so an agent following it hit an unhandled exception with nothing to act on.

## Fix

Pick the upgrader instead of assuming one. uv stays the first choice when it's on PATH, since `install.md` documents `uv tool install` as the blessed route; otherwise fall back to pip against the interpreter actually running us. When neither exists, say so and return 1.

Two smaller failure paths get handled while here:

- **`subprocess.run` wrapped in `try/except OSError`.** `which()` can resolve a binary that is gone or unexecutable by the time we spawn it. That must not surface as a traceback either.
- **A pip install on a box that happens to have uv.** `uv tool upgrade` only manages packages from `uv tool install`, so this case fails with `` `browser-harness` is not installed; run `uv tool install browser-harness` `` — accurate from uv's point of view, useless from the user's. It's now followed by the pip command that will actually work.

That second case is not hypothetical; it's what this checkout does. Forcing the `pypi` branch here (uv present, not a uv-tool install) gives exactly that dead end:

```
$ browser-harness --update -y
updating browser-harness: 0.1.9 -> 0.1.10
error: Failed to upgrade browser-harness
  Caused by: `browser-harness` is not installed; run `uv tool install browser-harness` to install
```

## Verification

The reported crash, reproduced by actually removing uv from `PATH` rather than by stubbing — `subprocess` resolves the binary itself, so monkeypatching `shutil.which` does **not** reproduce it:

```
$ PATH=C:\Windows\System32 python -c "...run_update(yes=True)"     # unfixed
  File "subprocess.py", line 1538, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args, ...)
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

Same scenario on this branch:

```
updating browser-harness: 0.1.9 -> 0.1.10
update complete.
returned: 0
ran     : python.exe ['-m', 'pip', 'install', '--upgrade', 'browser-harness']
```

Seven new tests — the three command-selection branches, the reported crash, the no-upgrader path, a stale PATH entry, and the pip hint on uv failure. All seven fail on `main` and pass here.

Full suite: `9 failed, 159 passed` → `9 failed, 166 passed`. The 9 are pre-existing and unrelated — 8 POSIX-only scaffolding in `test_admin.py` (#680) and `test_skill.py` (#678, fixed in #693).

## Notes for reviewers

- **Detection vs. fallback.** I considered sniffing whether this interpreter lives in a uv tools environment (`uv-receipt.toml`, or `uv`/`tools` path components) to pick the right command outright. I left it out: it depends on uv's on-disk layout, which isn't a stable public contract, and the actionable error message covers the same case without betting on internals. Happy to add it if you'd rather have the automatic route.
- **pipx** installs land on the pip branch. pipx venvs ship pip, so `python -m pip install --upgrade` works there; I didn't add a separate `pipx upgrade` branch since nothing in the repo documents pipx.
- The `git` branch is untouched.

---
## Summary by cubic
Fixes #688: `run_update()` no longer assumes `uv` is installed when upgrading a PyPI install, so `pip`- or `pipx`-installed users no longer hit a traceback on the daily `--update -y` banner. It tries `uv` first when on PATH, then falls back to `pip` on the running interpreter when uv is missing or cannot upgrade the install.

**Bug Fixes**
- Prints an actionable error and returns 1 when neither `uv` nor `pip` is available.
- Catches `OSError` from `subprocess.run` so a stale PATH entry falls through to the next candidate instead of crashing.
- When `uv tool upgrade` fails on a pip-installed package, retries with `python -m pip install --upgrade`; a uv-managed install has no pip in its venv, so uv's failure still stands there.
- Intermediate candidates report "could not upgrade...; retrying with..." and only the final failure is called "upgrade failed", so a run that succeeds never prints error language.

<sup>Written for commit 5338c317e719a9f4f0125f9ae1a56b4fa578020f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/702?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/browser_harness/admin.py
+++ b/src/browser_harness/admin.py
@@ -1194,6 +1194,29 @@ def _prompt_yes(question, default_yes=True, yes=False):
     return ans.startswith("y")
 
 
+def _pip_available():
+    import importlib.util
+
+    return importlib.util.find_spec("pip") is not None
+
+
+def _pypi_upgrade_command():
+    """Command that can upgrade a wheel install, or None when nothing can.
+
+    install.md documents `uv tool install`, so uv stays the first choice when it
+    is on PATH. It is not always there — browser-harness is a normal PyPI
+    package and pip and pipx install it just as well — so fall back to pip
+    against the interpreter that is actually running us.
+    """
+    import shutil
+
+    if shutil.which("uv"):
+        return ["uv", "tool", "upgrade", "browser-harness"]
+    if _pip_available():
+        return [sys.executable, "-m", "pip", "install", "--upgrade", "browser-harness"]
+    return None
+
+
 def run_update(yes=False):
     """Pull the latest version and (after prompt) restart the daemon so it picks up changed code.
 
@@ -1227,8 +1250,33 @@ def run_update(yes=False):
         if r.returncode != 0:
             return r.returncode
     elif mode == "pypi":
-        tool_upgrade = subprocess.run(["uv", "tool", "upgrade", "browser-harness"])
+        command = _pypi_upgrade_command()
+        if command is None:
+            print(
+                "can't auto-update: neither uv nor pip is available. Install uv "
+                "(https://docs.astral.sh/uv/) or reinstall browser-harness with pip.",
+                file=sys.stderr,
+            )
+            return 1
+        printable = " ".join(command)
+        try:
+            tool_upgrade = subprocess.run(command)
+        except OSError as e:
+            # shutil.which() said the binary was there; PATH can still be stale,
+            # or the file can be unexecutable. Never surface this as a traceback.
+            print(f"could not run `{printable}`: {e}", file=sys.stderr)
+            return 1
         if tool_upgrade.returncode != 0:
+            print(f"upgrade failed: `{printable}`", file=sys.stderr)
+            if command[0] == "uv":
+                # uv tool upgrade only manages packages from `uv tool install`,
+                # so this is the expected failure for a pip install on a box
+                # that happens to have uv.
+                print(
+                    "if you installed browser-harness with pip, upgrade with:\n"
+                    f"  {sys.executable} -m pip install --upgrade browser-harness",
+                    file=sys.stderr,
+                )
             return tool_upgrade.returncode
     else:
         print("unknown install mode; can't auto-update.", file=sys.stderr)
--- a/src/browser_harness/admin.py
+++ b/src/browser_harness/admin.py
@@ -1200,21 +1200,51 @@ def _pip_available():
     return importlib.util.find_spec("pip") is not None
 
 
-def _pypi_upgrade_command():
-    """Command that can upgrade a wheel install, or None when nothing can.
+def _pypi_upgrade_commands():
+    """Upgrade candidates for a wheel install, best first. Empty when none apply.
 
-    install.md documents `uv tool install`, so uv stays the first choice when it
-    is on PATH. It is not always there — browser-harness is a normal PyPI
-    package and pip and pipx install it just as well — so fall back to pip
-    against the interpreter that is actually running us.
+    uv leads whenever it is on PATH, because install.md documents
+    `uv tool install`. pip follows rather than replacing it: `uv tool upgrade`
+    only manages what `uv tool install` put there, so on a pip or pipx install
+    it fails with "`browser-harness` is not installed" — having uv on PATH must
+    not be the end of the road for those users.
+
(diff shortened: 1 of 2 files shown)


rajarshidattapy:
Valid, and a fair hit, the PR promised a fallback and delivered a suggestion. The pip path only ran when uv was *absent* from PATH; when uv was present but didn't manage the install, `--update` stopped at uv's failure and told the user to run pip themselves.

rajarshidattapy:
Valid, that line was in my own verification output in the previous comment and I read straight past it. `exit: 0` with `upgrade failed` on stderr.

PR #704, opened 2026-08-29
fix: stop doctor reporting FAIL for browsers the daemon drives

### LINKED ISSUE #686: `--doctor` reports "chrome running FAIL" and exits 1 for Brave and Chromium users, while the daemon works fine
**Labels:** bug, doctor, ux

## Description

Two functions answer the same question — "is a supported browser running?" — from two
different, inconsistent process-name lists.

The daemon's list is correct and includes everything in `PROFILES`:

```python
# src/browser_harness/daemon.py:179
return any(n in out for n in ("chrome.exe", "msedge.exe", "chromium.exe", "brave.exe", "helium.exe"))
```

Doctor's list is a strict subset, and drops Brave on both platforms and Chromium on
Windows:

```python
# src/browser_harness/admin.py:900-913
if system == "Windows":
    out = subprocess.check_output(["tasklist"], ...)
    names = ("chrome.exe", "msedge.exe", "helium.exe")            # no chromium.exe, no brave.exe
else:
    out = subprocess.check_output(["ps", "-A", "-o", "comm="], ...)
    names = ("Google Chrome", "chrome", "chromium", "Microsoft Edge", "msedge", "helium")  # no brave
```

Brave is a first-class target elsewhere in the codebase:

- `daemon._MAC_PROFILES` → `Library/Application Support/BraveSoftware/Brave-Browser`
- `daemon._LINUX_PROFILES` → `.config/BraveSoftware/...` (via the flatpak entry)
- `daemon._WINDOWS_PROFILES` → `BraveSoftware/Brave-Browser/User Data`
- `admin._BROWSER_LAUNCH` → `("brave", "Brave Browser", ("brave-browser", "brave"), "brave")`

So the harness will discover a Brave profile, connect to Brave over CDP, and drive it —
and then `--doctor` will tell the user Brave is not running.

## Impact

`run_doctor()`'s exit code is derived from this check:

```python
# admin.py:1152
return 0 if (chrome and daemon) else 1
```

A Brave user with a healthy daemon and a live browser connection gets:

```
  [FAIL] chrome running — start chrome/edge
  [ok  ] daemon alive
  [ok  ] active browser connections — 1
```

exit code 1. `install.md` and `SKILL.md` both direct agents to `--doctor` as *the*
diagnostic when something looks wrong, so the agent is handed a false failure and a
"start chrome/edge" instruction while its browser is already attached and working.
The row directly below it contradicting the FAIL makes it worse, not better.

Windows Chromium users hit the same thing (`chromium.exe` is missing from doctor's list
but present in the daemon's).

## Suggested fix

There should be one list, not two. `admin._chrome_running()` should call
`daemon.supported_browser_running()` — it already imports from `.daemon` in several
places — or both should read a shared constant. Either way, add `brave.exe`/`brave` and
`chromium.exe`.

Secondary: `run_doctor()` exiting 1 on the *process-name* check while
`active browser connections` is non-zero is backwards. A live CDP connection is stronger
evidence than a `tasklist` substring match, and should be able to satisfy the health
check on its own — `run_doctor_json()` already gets this right
(`healthy = browser_ready or (chrome and daemon)`, `admin.py:1166`).

rajarshidattapy:
Fixes #686.

## Problem

Two hand-maintained lists answer the same question — "is a supported browser running?" — and doctor's copy had drifted:

| | list |
|---|---|
| `daemon.supported_browser_running()` | `chrome.exe` `msedge.exe` `chromium.exe` `brave.exe` `helium.exe` |
| `admin._chrome_running()` (Windows) | `chrome.exe` `msedge.exe` — — `helium.exe` |
| `admin._chrome_running()` (POSIX) | `chrome` `chromium` `Microsoft Edge` `msedge` — `helium` |

Brave is missing from both of doctor's lists, Chromium from its Windows one. Yet Brave has a profile entry in all three of `_MAC_PROFILES`, `_LINUX_PROFILES` and `_WINDOWS_PROFILES`, and a launch spec in `_BROWSER_LAUNCH` — the harness discovers a Brave profile, attaches to it over CDP and drives it. Doctor then prints:

```
  [FAIL] chrome running — start chrome/edge
  [ok  ] daemon alive
  [ok  ] active browser connections — 1
```

and exits 1, because `run_doctor()` gates its exit code on that check. `SKILL.md` and `install.md` both send agents to `--doctor` when something looks wrong, so the agent gets a false failure telling it to start a browser that is already attached — with the very next line contradicting it.

## Fix

**One list.** Moved the process names to `daemon.py` as `BROWSER_PROCESSES_WINDOWS` / `BROWSER_PROCESSES_POSIX`, read from both call sites, so they cannot drift apart again. Brave and Chromium come along for free.

**A live CDP connection satisfies the check.** It is strictly better evidence than a process-name match, and it covers what the names cannot. Arc, Dia and Comet have `_MAC_PROFILES` entries but are genuinely unsafe to substring-match:

```
'dia'   -> matches 'mediaremoted', 'com.apple.MediaRemote', 'diagnosticd'
'arc'   -> matches 'searchd', 'arcd'
'comet' -> matches 'Comet Helper'
```

Listing them would make doctor claim a browser is running on any idle Mac, so they stay out. macOS already finds them via `SingletonLock` in the daemon, and doctor now falls back to the connection — which is also why I did **not** simply make `_chrome_running()` call `supported_browser_running()`: that function's POSIX branch checks `SingletonLock` in the standard profile dirs only, so it would report False for a dedicated automation Chrome started with a custom `--user-data-dir` under `BU_CDP_URL`, which `ps` currently catches.

`run_doctor_json()` already scored `browser_ready` ahead of the name probe (`browser_ready or (chrome and daemon)`). `run_doctor()` now agrees with it.

## Verification

Nine new tests; six fail on `main`:

```
FAILED test_chrome_running_detects_brave[windows]
FAILED test_chrome_running_detects_brave[linux]
FAILED test_chrome_running_detects_brave[darwin]
FAILED test_chrome_running_detects_chromium_on_windows
FAILED test_doctor_and_daemon_agree_on_windows_process_names
FAILED test_doctor_trusts_a_live_connection_over_the_name_probe
6 failed, 3 passed
```

The 3 that pass both ways are the negative guards — no browser, and no-browser-plus-no-connection still fails — there to stop the fix from turning doctor into something that always reports OK.

One note on how those were built: my first Windows fixture listed `brave.exe` *alongside* `chrome.exe`, so it passed against the unfixed code for the wrong reason. The fixtures now contain only the browser under test (`explorer.exe` as filler), which is what moved `[windows]` from passing to failing on `main`.

`test_doctor_and_daemon_agree_on_windows_process_names` iterates `daemon.BROWSER_PROCESSES_WINDOWS` and asserts doctor detects each one, so the sharing is pinned behaviourally rather than by identity — re-introducing a separate literal list in `admin.py` fails it.

Full suite: `9 failed, 159 passed` → `9 failed, 167 passed`. The 9 are pre-existing and unrelated (#680, #678).

## Note for reviewers

The POSIX detection methods still differ by design — the daemon checks `SingletonLock` per profile dir (precise, and what its recovery logic needs), doctor greps `ps` (broad, and what a diagnostic wants). Only the name lists are shared. Worth a look on a real Brave/Linux box; everything here was verified on Windows with stubbed process output.

---
## Summary by cubic
Fixes #686 so `--doctor` and `doctor --json` trust the daemon's browser process list and treat a live CDP connection as proof a browser is running.

- Moves the browser process names to `daemon.py` (`BROWSER_PROCESSES_WINDOWS` / `BROWSER_PROCESSES_POSIX`) and reads them from both call sites, so they cannot drift apart.
- Lets a live CDP connection satisfy the check before the name probe, covering Arc, Dia, and Comet without unsafe name matching.
- `--doctor` and `doctor --json` now exit 0 for these cases; strict JSON mode still reports `null` for `chrome_running`.

<sup>Written for commit 1d1c67245d61c795ebf4b5cd6a8613f2adb1e471. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/704?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/browser_harness/admin.py
+++ b/src/browser_harness/admin.py
@@ -898,16 +898,24 @@ def print_update_banner(out=None):
 
 
 def _chrome_running():
-    """Cross-platform best-effort check for a running Chromium-based browser."""
+    """Cross-platform best-effort check for a running Chromium-based browser.
+
+    Shares its process names with daemon.supported_browser_running(). They used
+    to be two hand-maintained lists, and doctor's had drifted: it was missing
+    Brave on both platforms and Chromium on Windows, so users of browsers the
+    harness happily drives were told to "start chrome/edge".
+    """
     import platform, subprocess
+    from .daemon import BROWSER_PROCESSES_POSIX, BROWSER_PROCESSES_WINDOWS
+
     system = platform.system()
     try:
         if system == "Windows":
             out = subprocess.check_output(["tasklist"], text=True, errors="replace", timeout=5)
-            names = ("chrome.exe", "msedge.exe", "helium.exe")
+            names = BROWSER_PROCESSES_WINDOWS
         else:
             out = subprocess.check_output(["ps", "-A", "-o", "comm="], text=True, errors="replace", timeout=5)
-            names = ("Google Chrome", "chrome", "chromium", "Microsoft Edge", "msedge", "helium")
+            names = BROWSER_PROCESSES_POSIX
         return any(n.lower() in out.lower() for n in names)
     except Exception:
         return False
@@ -1102,9 +1110,14 @@ def run_doctor():
     import platform, sys
     cur = _version()
     mode = _install_mode()
-    chrome = _chrome_running()
     daemon = daemon_alive()
     connections = browser_connections()
+    # A live CDP connection proves a browser is running, and proves it better
+    # than a process-name match: it covers the browsers the name probe cannot
+    # safely spot (Arc, Dia, Comet) and a Chrome launched from an unusual path.
+    # Without this, doctor could print a FAIL directly above the attached page
+    # it had just listed. run_doctor_json() already scores browser_ready first.
+    chrome = _chrome_running() or bool(connections)
     try:
         auth_state = auth.auth_status()
     except (auth.AuthError, OSError) as e:
--- a/src/browser_harness/admin.py
+++ b/src/browser_harness/admin.py
@@ -1105,6 +1105,21 @@ def _open_chrome_inspect_once():
         pass
 
 
+def _browser_running(live_connection):
+    """Is a browser running? A live CDP connection settles it on its own.
+
+    Better evidence than a process-name match: it covers the browsers the name
+    probe cannot safely spot (Arc, Dia and Comet have profile dirs but names too
+    common to substring-match — "arc" matches "searchd") and a Chrome launched
+    from an unusual path. Without it, `--doctor` could print a FAIL directly
+    above the attached page it had just listed.
+
+    Shared by both doctor surfaces so `--doctor` and `doctor --json` cannot
+    disagree about what "chrome running" means on the same machine.
+    """
+    return _chrome_running() or bool(live_connection)
+
+
 def run_doctor():
     """Read-only diagnostics. Exit 0 iff everything looks healthy."""
     import platform, sys
@@ -1112,12 +1127,7 @@ def run_doctor():
     mode = _install_mode()
     daemon = daemon_alive()
     connections = browser_connections()
-    # A live CDP connection proves a browser is running, and proves it better
-    # than a process-name match: it covers the browsers the name probe cannot
-    # safely spot (Arc, Dia, Comet) and a Chrome launched from an unusual path.
-    # Without this, doctor could print a FAIL directly above the attached page
-    # it had just listed. run_doctor_json() already scores browser_ready first.
-    chrome = _chrome_running() or bool(connections)
+    chrome = _browser_running(connections)
     try:
         auth_state = auth.auth_status()
     except (auth.AuthError, OSError) as e:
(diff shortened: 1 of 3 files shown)


rajarshidattapy:
Valid, I'd put the rule at run_doctor()'s call site instead of in a shared definition, so `doctor --json` still reported `chrome_running: false` where the text doctor said ok. Extracted it to `_browser_running()`, now used by both surfaces (strict mode still reports null, and `healthy` is unchanged since `browser_ready` already short-circuited it). Pushed as `a017564`.

PR #735, opened 2026-09-03
fix: fail closed in get_ws_url() when a named daemon opts out of local fallback

### LINKED ISSUE #479: get_ws_url() should fail CLOSED for a named/managed daemon instead of scanning the default Chrome profile
## Summary

`get_ws_url()` (`src/browser_harness/daemon.py:104`) falls back to scanning the **default** Chrome profiles and then probing ports 9222/9223 whenever `BU_CDP_URL`/`BU_CDP_WS` is unset. For a **named/managed** daemon (one started via `ensure_daemon(name=...)` to drive a *dedicated, isolated* automation Chrome), that fallback is a footgun: on any daemon respawn where the env doesn't carry `BU_CDP_URL`, the daemon silently attaches to the user's **daily-driver Chrome** — or pops the `chrome://inspect` "Allow remote debugging" dialog on Chrome 136+.

## Why it matters

A managed caller sets `BU_NAME` + `BU_CDP_URL` to pin a daemon to an isolated Chromium. But `ensure_daemon()` re-runs on **every** helper invocation, and on a respawn (parent restart, session teardown, zombie Chromium) a caller that carried only `BU_NAME` loses the pin. `get_ws_url()` then:

1. `daemon.py:126` — scans `PROFILES` (the **default** user-data-dirs), attaching to the user's real Chrome;
2. `daemon.py:154` — probes `9222`/`9223`, silently attaching to whatever is listening there.

Today this only *appears* safe because Chrome 136+ gates the default profile behind the "Allow" dialog — i.e. it fails closed by accident. If the user has a pre-authorized debug Chrome on 9222, it drives it silently. This bit us as a real cross-project daily-driver breach (a downstream broker's ADR-022 firewall).

## Proposed fix

An opt-in **fail-closed** guard so a managed daemon never falls back to local discovery. Either:

- honor a `BU_NO_LOCAL_FALLBACK=1` env, or
- infer it when the daemon was started with a `name` (managed daemons are, by definition, pinned).

When set, and `BU_CDP_URL`/`BU_CDP_WS` is absent or unreachable, `get_ws_url()` should `raise` rather than reach the `PROFILES` scan (`daemon.py:126`) or the `9222/9223` probe (`daemon.py:154`) — a loud, honest failure that says "the pinned endpoint is gone, relaunch the session," never a silent attach to the user's Chrome.

## Related

Distinct from #291 (that was a `/json/version` 404 fallback for the `BU_CDP_URL` *happy* path). This is about refusing the *default-profile* fallback entirely for pinned daemons.

dean985:
get_ws_url() falls back to scanning the default Chrome and Brave profile dirs whenever BU_CDP_WS and BU_CDP_URL are both unset. That includes a named daemon that was pinned to a dedicated automation Chrome. If that daemon respawns and loses its pinned env, it silently attaches to whatever browser happens to be running on the user's own daily driver profile instead of failing.

Named daemons also have a real, existing setup where several of them intentionally share the user's default local browser, each getting its own tab. A blanket rule that named daemons always refuse the fallback would break that, so the guard is opt in through BU_NO_LOCAL_FALLBACK=1 rather than automatic.

Adds two tests in tests/unit/test_daemon.py: one confirming a named daemon with the flag set refuses to scan the default profile, one confirming a named daemon without the flag still works as before.

Fixes #479

---
## Summary by cubic
Fixes `get_ws_url()` so a named daemon that loses its pinned `BU_CDP_URL`/`BU_CDP_WS` endpoint fails loudly instead of silently attaching to whatever browser is running on the user's default Chrome/Brave profile.

- The guard is opt-in via `BU_NO_LOCAL_FALLBACK=1`; named daemons that intentionally share the user's default browser keep scanning profiles as before.
- `ensure_daemon()` now lets its explicit `name` argument override any `BU_NAME` in a caller-supplied `env`, so callers can't bypass the fail-closed guard by sneaking in `BU_NAME=default`.
- Adds regression tests for the fail-closed path, the unchanged default fallback, and the `name`-over-`env` precedence.

Fixes #479.

<sup>Written for commit 6bd8f230342a6e1d0b449c0d9700dabf247a6e80. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/735?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/browser_harness/admin.py
+++ b/src/browser_harness/admin.py
@@ -370,7 +370,11 @@ def ensure_daemon(wait=60.0, name=None, env=None):
     launched_browser = None
     opened_inspect = False
     for _ in range(3):
-        e = {**os.environ, **({"BU_NAME": name} if name else {}), **(env or {})}
+        # The explicit `name` argument must win over anything in `env` — otherwise a
+        # caller-supplied env={"BU_NAME": ...} silently overrides which daemon this
+        # spawns as, defeating callers (and the fail-closed guard in get_ws_url())
+        # that rely on `name` to say what the daemon's identity actually is.
+        e = {**os.environ, **(env or {}), **({"BU_NAME": name} if name else {})}
         try:
             stderr_sink = open(ipc.log_path(name or NAME), "ab")
         except OSError:
--- a/src/browser_harness/daemon.py
+++ b/src/browser_harness/daemon.py
@@ -252,6 +252,20 @@ def get_ws_url():
         if platform.system() == "Windows":
             hint += "; on Windows also check that a firewall/antivirus isn't blocking localhost connections"
         raise RuntimeError(f"BU_CDP_URL={url} unreachable after 30s: {last_err} -- {hint}")
+    # A named/managed daemon may have been launched with a pinned BU_CDP_URL that a
+    # later respawn (parent restart, session teardown) fails to carry forward. Falling
+    # through to the local profile scan below would then silently attach to whatever
+    # browser happens to be running on the user's default profile instead of the
+    # dedicated automation Chrome the caller pinned -- opt in to failing loudly instead
+    # with BU_NO_LOCAL_FALLBACK=1. Off by default: named local daemons sharing the
+    # user's default browser (each getting its own tab, see attach_first_page) are a
+    # legitimate, currently-supported setup and must keep scanning PROFILES normally.
+    if NAME != "default" and (os.environ.get("BU_NO_LOCAL_FALLBACK") or "").strip().lower() not in ("", "0", "false", "no", "off"):
+        raise RuntimeError(
+            f"pinned endpoint lost for managed daemon BU_NAME={NAME!r}: BU_CDP_WS/BU_CDP_URL "
+            "is unset and BU_NO_LOCAL_FALLBACK is set, so refusing to scan the default browser "
+            "profile -- relaunch the session with its pinned endpoint restored"
+        )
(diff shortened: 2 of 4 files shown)


MagMueller:
found one bypass: with BU_NO_LOCAL_FALLBACK=1, ensure_daemon(name="managed", env={"BU_NAME": "default"}) still discovers the default profile. putting the explicit name last in admin.py's env merge fixes it. verified with real child processes and a synthetic CDP endpoint, not Chrome. could you include that?

dean985:
Good catches, both fixed.

@MagMueller — `ensure_daemon` merged `env` after the explicit `name`, so `env={"BU_NAME": "default"}` overrode `name="managed"`, which skips the new guard (it only fires when `NAME != "default"`). Swapped the merge order, added a test with your exact repro that fails on the old order and passes now.

cubic — the companion test's fake `urlopen` accepted any URL, so it wouldn't have caught skipping the `PROFILES` scan. Now asserts the request actually hit the `DevToolsActivePort` port.

Pushed both, 224 tests green.

PR #736, opened 2026-09-03
Tell pip and pipx users how to upgrade when uv tool upgrade fails

MagMueller:
`browser-harness --update` on a wheel install runs `uv tool upgrade browser-harness`. That command only manages installs created by `uv tool install`, so a pip or pipx install fails with ``​`browser-harness` is not installed`` and the user is left with a bare non-zero exit and no next step.

This prints the documented uv install command on that failure:

```
if you installed with pip or pipx, upgrade with: uv tool install --python 3.12 --upgrade --force browser-harness
```

The command matches `install.md`. Behaviour is unchanged otherwise: the non-zero return code from `uv tool upgrade` is still returned, and nothing new runs.

Tests: `pytest tests -q` → 244 passed. `git diff --check` clean.

https://claude.ai/code/session_01UV3AvcViJQzAV75u67XZCF

---
## Summary by cubic
When `uv tool upgrade` fails, `browser-harness --update` now checks `uv tool list` and only shows the pip/pipx upgrade hint when `uv` does not manage the install. The entry name is matched exactly, so a similarly named tool won't silence the hint; uv-managed failures and unknown ownership keep their original output and non-zero exit code.

<sup>Written for commit 0c1d09a45e24a342455cf1afb9b577e49d91eb5a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/736?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/browser_harness/admin.py
+++ b/src/browser_harness/admin.py
@@ -1216,7 +1216,16 @@ def _uv_manages_browser_harness():
         return True
     if listed.returncode != 0:
         return True
-    return "browser-harness" in (listed.stdout or "")
+    # `uv tool list` prints one "name vX.Y.Z" line per tool, then its executables as
+    # "- exe" lines. Match the entry name, so a tool merely containing our name (say
+    # my-browser-harness-wrapper) cannot silence the hint for a real pip install.
+    for line in (listed.stdout or "").splitlines():
+        entry = line.strip()
+        if entry.startswith("-"):
+            continue
+        if entry.split(" ", 1)[0] == "browser-harness":
+            return True
+    return False
 
 
 def run_update(yes=False):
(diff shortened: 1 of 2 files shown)


MagMueller:
Good catch, fixed in f533f9e. The hint now checks `uv tool list` first and only prints when browser-harness is absent, so a uv-managed install that fails for another reason (offline, auth) gets uv's own message and nothing else. Unknown counts as managed, so a missing or broken uv stays quiet too. Two tests cover both branches; the quiet one fails on the previous commit.

Issue #631, opened 2026-08-20
Repeated 'Allow remote debugging?' approvals: daemon churn defeats the one-click-per-Chrome-run design


johnblythe:
### What happens

On Chrome 144+ every new CDP connection costs one "Allow remote debugging?" approval. The daemon's single held connection is meant to make that a one-time click per Chrome run, but in practice users get asked over and over. Four interacting bugs churn the connection:

1. **`ensure_daemon` kills the daemon that is holding the popup.** When `wait` expires (or the daemon's own 45s handshake window runs out), `ensure_daemon` calls `restart_daemon(name)` and raises "retry". Each retry spawns a new daemon, which opens a new WS handshake, which raises a new popup. Worse, `restart_daemon` cannot even see a parked daemon (it has no IPC socket yet, so `identify()`/`ping()` both fail), so it only unlinks the pid file: the parked daemon survives as an orphan holding a now-stale popup while its replacement raises a second one.

2. **45 seconds of popup patience.** `LOCAL_HANDSHAKE_TIMEOUT = 45` means a user who is not looking at Chrome (other app, other desktop, agent running in the background) loses the popup before they ever see it, guaranteeing a re-pop on the next attempt.

3. **Concurrent invocations stack popups.** A parked daemon cannot answer pings, so a second `browser-harness` invocation (parallel agent session, or an agent retrying) passes `already_running()` and spawns a sibling daemon: a second popup, plus the sibling truncates the first daemon's log (destroying the `handshake-wait` breadcrumb `ensure_daemon` keys on) and overwrites its pid file.

4. **Hair-trigger staleness probe.** A live daemon gets two `Target.getTargets` probes with 3s timeouts before being killed and respawned. A busy Chrome (heavy page, memory pressure) trips this, and the pointless restart costs an approval.

`mac-approve` (added in 0.1.9) is a great mitigation on macOS with Accessibility granted, but it does not remove the churn: non-macOS users and macOS users without Accessibility still click every popup, and even the auto-approver ends up chasing popups that the churn keeps dismissing and re-raising.

### Repro sketches

- Run any `browser-harness` command with remote debugging enabled but do not click Allow within ~60s. The command fails, the popup vanishes, and the next attempt raises a fresh popup. Repeat forever.
- Run two `browser-harness` commands near-simultaneously on a cold daemon: two popups.
- Put Chrome under heavy load so a `Target.getTargets` round-trip exceeds 3s twice: healthy daemon restarted, one popup for nothing.

[...]

PR #737, opened 2026-09-03
Keep one Chrome approval pending per daemon

MagMueller:
Chrome 144+ asks for approval for every new CDP connection. Browser Harness is meant to hold one connection open, but three startup paths could turn one pending approval into several prompts for the same daemon name:

1. `ensure_daemon` killed the process holding the permission sheet after its caller stopped waiting.
2. A daemon has no IPC socket until Chrome accepts the connection, so concurrent cold callers could start sibling processes.
3. The WebSocket handshake expired after 45 seconds.

This keeps the first pending daemon alive, atomically publishes a process-fingerprinted PID while holding a per-name owner lock, and makes later callers join it instead of opening another Chrome connection. The lock cannot be age-stolen from a live owner after laptop sleep, and an old owner cannot unlink a successor lock.

For local Chrome, the approval handshake and the original default caller have no deadline. They wait until Chrome approves, Chrome or the daemon exits, the transport actually fails, or the user explicitly runs `--reload`. Approval failures never auto-retry or create a replacement connection. `--reload` serializes its ownership snapshot, cancellation, and cleanup under the same per-name lock, then signals only the exact atomically published PID whose process-start fingerprint still matches. Stale, legacy, PID-reused, and successor records are never trusted.

Browser Use Cloud and explicit `BU_CDP_URL` / `BU_CDP_WS` startup retain their existing bounded behavior. The no-deadline path is local-Chrome-only.

On macOS the bundled skill tells the agent to keep the original command running and invoke `mac-approve` once with the exact same `BU_NAME`. It explains the one-time Accessibility grant and explicitly excludes remote and cloud browsers.

After approval, the daemon stays alive until Chrome closes the CDP session, the machine or process kills it, or Browser Harness is explicitly reloaded.

## Proof

- Full unit suite: 268 passed.
- Deterministic regressions cover atomic daemon PID publication, live-lock survival across old mtimes, successor-generation ownership, PID reuse, fail-closed process identity, deadline-free local approval, bounded remote startup, no retry after an approval failure, and fingerprint-verified cancellation of a pre-IPC pending daemon.
- Live macOS Chrome on the final no-deadline code: a fresh named daemon raised one sheet and printed the exact-name helper command; `mac-approve` returned `ready`; the original waiting command resumed and loaded Hacker News; a second command reused the connection without another prompt. The test tabs and daemon were then closed.
- Live cancellation: another pending named daemon was stopped with `--reload`; the waiting caller exited without retrying, and independent checks confirmed its process, PID record, and endpoint were gone.
- Chrome's M144 documentation says it asks every time a client requests a remote debugging session.

## Scope boundary

Different `BU_NAME` values still create different CDP connections. Sharing one daemon among concurrent agents requires explicit tab/session coordination because the daemon has shared mutable default-tab state. The existing staleness probe from #631 is also not changed here. This PR prevents duplicate pending starts and approval retry churn; it does not claim one connection across all names or all of Chrome's lifetime.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -61,14 +61,22 @@ If Chrome is running but remote debugging is not enabled, the harness opens:
 chrome://inspect/#remote-debugging
 ```
 
-On macOS, when Chrome asks for remote-debugging permission, run:
+On macOS, when local Chrome asks for remote-debugging permission, keep the
+original browser command running and call `mac-approve` in another shell/tool
+call. Preserve the exact daemon name: if the waiting command used
+`BU_NAME=r7k2`, run:
 
 ```text
-browser-harness mac-approve
+BU_NAME=r7k2 browser-harness mac-approve
 ```
 
-Continue browser work when it returns `ready`; otherwise follow its printed
-instruction.
+For the default daemon, omit the `BU_NAME` prefix. The original command resumes
+when the helper returns `ready`; do not rerun it. If the helper reports
+`accessibility-required`, ask the user once to grant the app launching
+browser-harness (for example Terminal, iTerm, or Codex) access in System
+Settings > Privacy & Security > Accessibility, then call `mac-approve` once
+again. This is only for local Chrome; do not call it for `BU_CDP_URL`,
+`BU_CDP_WS`, or Browser Use Cloud.
 
 ## Remote Browsers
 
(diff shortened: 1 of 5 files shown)


MagMueller:
Fixed all three findings.

- **pid reuse (P2):** `_parked_daemon_pid` now requires the `handshake-wait` breadcrumb to be fresh (`_PARKED_LOG_GRACE`, 120s) *and* the pid to look like one of our daemons via `ps -o command=`. Where that check cannot run (Windows, no `ps`) it falls back to the freshness guard rather than blocking startup. Two new tests cover the recycled pid and the stale breadcrumb.
- **concurrent spawn (P2):** detection and spawning now run inside `_spawn_lock`, a best-effort cross-process lock (`O_CREAT|O_EXCL`) with its own stale clock, separate from the caller's wait timeout. `daemon_alive` and the parked check are re-run inside the lock. If the lock cannot be taken we still spawn, because failing to start is worse than an extra popup. Covered by `test_spawn_lock_is_exclusive_then_released`.
- **dead hasattr branch (P3):** the test patches `admin_mod.subprocess.Popen` directly.

Full `tests/unit`: 236 passed.

PR #745, opened 2026-09-04
feat: multiplex agents across browser tabs

MagMueller:
## Summary

- make the default daemon own and reuse one CDP session per tab target
- have each CLI script remember only its selected `target_id`; all helper calls carry that target automatically
- route network events and JavaScript dialogs by target so agents on different tabs do not consume or observe each other's state
- recover a stale target session by reattaching to that exact target, never the daemon's unrelated default tab
- simplify the bundled skill: one default daemon, one tab per concurrent task, no agent IDs, context names, locks, or named local daemons

## Why

Chrome's browser WebSocket already multiplexes multiple target sessions. The harness was defeating that capability because every `switch_tab()` overwrote one daemon-global session. Two agents could therefore redirect each other's clicks, screenshots, and page reads even when they had different tabs.

The agent does not need to know about sessions or other agents. It creates/selects a tab, remembers that target ID between CLI invocations, and starts later scripts with `switch_tab(target_id)`. The daemon keeps the one browser-level connection alive and handles target-to-session mapping, event/dialog routing, and exact-target stale recovery internally.

Browser-wide `Browser.*` and `Target.*` commands bypass tab sessions. Two agents acting on the same tab are still a semantic race. Different tabs may operate concurrently through the same daemon. Remote browsers remain useful when cookies, storage, browser lifecycle, network, or crash isolation is required.

## Scope

Five files: the bundled skill, two runtime files, and their two unit-test files. No CLI lifecycle hook or persistent agent-context registry.

## Validation

- `uv run --with pytest python -m pytest tests/unit -q` (263 passed)
- `uv build --out-dir /tmp/browser-harness-multiplex-build-v2`
- `git diff --check`

A live local-Chrome smoke was intentionally not run from this worktree because it would require recycling or creating a browser-level daemon connection and could trigger the exact extra Chrome approval prompt this change is meant to avoid.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -25,9 +25,9 @@ PY
 
 - Invoke as `browser-harness`. Use heredocs for multi-line commands.
 - Helpers are pre-imported. `run.py` calls `ensure_daemon()` before `exec`.
-- First navigation for a task is `new_tab(url)`, not `goto_url(url)`. The daemon
-  preserves the attached tab across separate CLI invocations, so do not call
-  `new_tab()` again in every script.
+- First navigation for a task is `target_id = new_tab(url)`, not
+  `goto_url(url)`. Print and remember that target ID. At the start of each later
+  CLI invocation, call `switch_tab(target_id)`; do not call `new_tab()` again.
 - Keep one working tab per task/site. Before opening another, inspect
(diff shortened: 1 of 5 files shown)


MagMueller:
Closing in favor of the existing skill-only operating model: reuse one default local daemon and serialize local-browser operations. No runtime change requested.

PR #748, opened 2026-09-04
fix(helpers): switch to surviving tab before closing current tab (#379)

### LINKED ISSUE #379: close_tab() on the current tab leaves dangling session state

warun7:
## Problem Context

`close_tab()` closes the currently attached tab when called without arguments. It previously sent `Target.closeTarget` directly to Chrome without updating daemon session state.

The daemon still pointed to the closed `targetId` and `sessionId`. Subsequent calls such as `current_tab()`, `page_info()`, or `js()` failed immediately with `RuntimeError: cdp_disconnected`.

## Why This Fix Is Needed

`SKILL.md` instructs agents to close task tabs on completion. Calling `close_tab()` as documented left the daemon in a broken state for any subsequent interaction.

Callers had to manually call `switch_tab()` after closing to recover.

## Changes

- Detects when the target being closed is the currently attached tab.
- Switches to an available surviving user tab before issuing `Target.closeTarget`.
- Falls back to internal tabs or creates a fresh `about:blank` tab if no other user tab exists.
- Leaves closing of non-current tabs unchanged and fast.
- Adds comprehensive unit tests in `tests/unit/test_helpers.py`.

Fixes #379

---
## Summary by cubic
`close_tab()` now switches to a surviving tab before closing the currently attached tab, so the daemon keeps a valid session and later calls like `current_tab()`, `page_info()`, or `js()` still work. Previously it sent `Target.closeTarget` directly and left the daemon pointing at the closed tab, forcing callers to run `switch_tab()` manually to recover.

- Switching and closing are serialized under the session-state lock so concurrent `close_tab` and attach calls can't race.
- If another user tab exists, it switches to that; otherwise it reuses a Chrome internal tab or creates `about:blank`.
- Closing a non-current tab is unchanged and still closes directly.
- A failed `Target.closeTarget` (or a missing attached tab) now returns an error instead of silently leaving stale state.
- Adds unit tests for all close paths, including no-attached-tab behavior.

Fixes #379.

<sup>Written for commit 236dc8579833e7754a585012451495db6678bcc0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/748?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/browser_harness/daemon.py
+++ b/src/browser_harness/daemon.py
@@ -439,7 +439,7 @@ class Daemon:
         self.dialog = None
         self.stop = None  # asyncio.Event, set inside start()
 
-    async def attach_first_page(self, replaces_session=None, enable_domains=True):
+    async def attach_first_page(self, replaces_session=None, enable_domains=True, exclude_target_id=None):
         """Attach to a real page (or any page). Sets self.session. Returns attached target or None."""
         targets = (await self.cdp.send_raw("Target.getTargets"))["targetInfos"]
         # Named daemons (BU_NAME != "default") share one browser with other
@@ -452,7 +452,7 @@ class Daemon:
             # Clean it up before returning from this early path as well.
             if BROWSER_KIND == "local":
                 await self._close_inspect_tabs(targets)
-            pages_by_id = {t["targetId"]: t for t in targets if t["type"] == "page"}
+            pages_by_id = {t["targetId"]: t for t in targets if t["type"] == "page" and t["targetId"] != exclude_target_id}
             # A stale CDP session does not necessarily mean its tab disappeared.
             # Reattach to the current tab first, then the daemon's dedicated tab.
             page = pages_by_id.get(self.target_id) or pages_by_id.get(self.dedicated_target_id)
@@ -461,7 +461,7 @@ class Daemon:
                 # inside a narrow lock so they share one replacement tab.
                 async with self._dedicated_target_lock:
                     refreshed = (await self.cdp.send_raw("Target.getTargets"))["targetInfos"]
-                    pages_by_id = {t["targetId"]: t for t in refreshed if t["type"] == "page"}
+                    pages_by_id = {t["targetId"]: t for t in refreshed if t["type"] == "page" and t["targetId"] != exclude_target_id}
                     page = pages_by_id.get(self.target_id) or pages_by_id.get(self.dedicated_target_id)
                     if page is None:
                         tid = (await self.cdp.send_raw(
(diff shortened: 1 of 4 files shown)


warun7:
Serialized target selection, replacement attach, and Target.closeTarget in a daemon-side close_tab handler guarded by _session_state_lock. This prevents race conditions when closing tabs concurrently.

warun7:
Passed replaces_session=None when closing the current tab so in-flight requests are not redirected to an unrelated tab. Validated Target.closeTarget return status and treat false as an error.

PR #750, opened 2026-09-04
fix(auth): prevent OAuth callback server hang on browser preconnect (#691)

### LINKED ISSUE #691: `auth login`'s OAuth callback server can hang past its own timeout on a browser preconnect
**Labels:** bug, auth

## Description

The browser login flow serves its OAuth redirect from a single-threaded stdlib
`HTTPServer` on the main thread:

```python
# src/browser_harness/auth.py:421
return HTTPServer(("127.0.0.1", 0), Handler)
```

```python
# src/browser_harness/auth.py:249-256
def complete_browser_auth(start, *, timeout=AUTH_TIMEOUT_SECONDS):
    deadline = time.time() + timeout
    start.server.timeout = 0.5
    try:
        while not start.callback.complete and time.time() < deadline:
            start.server.handle_request()
```

`HTTPServer.timeout = 0.5` bounds only the `select()` *before* `accept()`. Once a
connection is accepted, `handle_request()` calls into the handler, and
`BaseHTTPRequestHandler.timeout` is never set — so the socket has no timeout and
`self.rfile.readline()` blocks until the peer sends a request line or closes.

## Impact

Browsers routinely open speculative TCP connections that carry no request. Chrome
preconnects to the origin as soon as a navigation to `http://127.0.0.1:<port>/...` is
predicted, and typically opens more sockets than it uses. If `accept()` returns one of
those before the real request arrives, `handle_request()` parks on a socket that will
never speak, and:

- the `time.time() < deadline` check is never reached again, so the 600-second
  `AUTH_TIMEOUT_SECONDS` bound does not apply;
- the real callback lands in the listen backlog and is never accepted;
- `browser-harness auth login` hangs with `Waiting for login to complete...` and no
  error, until the browser eventually drops the idle socket (which can be minutes) or
  the user gives up.

The failure is intermittent and timing-dependent, which makes it look like a flaky
server-side auth problem rather than a client bug.

## Suggested fix

Two one-line changes, either of which closes the hole; both together is better:

```python
class Handler(BaseHTTPRequestHandler):
    timeout = 10          # bounds readline() on an accepted socket
    ...

return ThreadingHTTPServer(("127.0.0.1", 0), Handler)   # a silent socket can't block the next accept
```

`ThreadingHTTPServer` is in `http.server` alongside the import already there, and
`PendingCallback` is only written from the handler and read from the loop, so no extra
locking is needed for the existing fields.

## Secondary, same file

`do_GET` returns `Browser Use Cloud login complete` for every request that reaches
`CALLBACK_PATH`, including `invalid_state` and provider-error callbacks
(`auth.py:400-417`). The user is told in the browser that login worked while the CLI
prints `auth failed: invalid_state`. The body should reflect `callback.error`.

warun7:
## Problem Context

The browser login flow serves its OAuth redirect from a single-threaded `HTTPServer` on the main thread.

Modern browsers open speculative preconnect TCP connections that carry no immediate request data. When `handle_request()` accepted an idle socket, `readline()` blocked indefinitely because `BaseHTTPRequestHandler.timeout` was not configured. This bypassed the 600-second deadline in `complete_browser_auth()`, leaving `browser-harness auth login` permanently hung.

Additionally, `do_GET` returned "Browser Use Cloud login complete" even when the state check failed or when the provider sent an OAuth error parameter.

## Why This Fix Is Needed

Users and unattended agents attempting `browser-harness auth login` experienced intermittent hangs.

Showing a success page on an authentication failure also misled users while the CLI reported failure.

## Changes

- Uses `ThreadingHTTPServer` with `daemon_threads = True` so idle sockets do not block incoming requests or process shutdown.
- Configures `timeout = 10` on the request handler to bound socket reads.
- Renders an error notice in HTML when `callback.error` is present.
- Sets `callback.complete = True` after the HTTP response is written.
- Adds `tests/unit/test_auth.py` covering successful OAuth, invalid state, provider errors, preconnect non-blocking behavior, and timeout conditions.

Fixes #691

---
## Summary by cubic
Prevents the OAuth callback server from hanging when a browser preconnect leaves an idle socket blocking the single-threaded server, and makes failures render an error page instead of a success page. This resolves the `browser-harness auth login` hang from #691.

**Bug Fixes**
- Switches to `ThreadingHTTPServer` with daemon threads and a 10-second socket read timeout.
- Serializes callback handling with a lock, validates that a code is present, and marks the callback complete only after the response is flushed.
- Renders an error notice in HTML when the callback contains an error.
- Adds unit tests covering successful OAuth, invalid state, provider errors, missing codes, duplicate requests, preconnect behavior, and timeouts.

<sup>Written for commit 8bc2019954615bdc1af65f5668ac15a0d5771f30. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/750?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/browser_harness/auth.py
+++ b/src/browser_harness/auth.py
@@ -6,11 +6,12 @@ or tells the agent to run `browser-harness auth login`. OAuth details live here.
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from http.server import BaseHTTPRequestHandler, HTTPServer
+from http.server import BaseHTTPRequestHandler, HTTPServer, ThreadingHTTPServer
 import argparse
 import base64
 import getpass
 import hashlib
+import html
 import json
 import os
 from pathlib import Path
@@ -393,6 +394,8 @@ def _exchange_authorization_code(code: str, redirect_uri: str, verifier: str) ->
 
 def _callback_server(callback: PendingCallback) -> HTTPServer:
     class Handler(BaseHTTPRequestHandler):
+        timeout = 10
+
         def do_GET(self):  # noqa: N802 - stdlib handler API
             parsed = urllib.parse.urlparse(self.path)
             if parsed.path != CALLBACK_PATH:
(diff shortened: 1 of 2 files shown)


warun7:
Serialized callback processing with a threading lock to prevent concurrent mutations. Missing authorization codes now trigger a callback failure response.

warun7:
Wrote and flushed the HTTP response before marking callback.complete under the same lock. This guarantees the browser receives the response before the server can be closed.

PR #811, opened 2026-09-18
fix(windows): replace AF_UNIX with cross-platform ipc module

LiewSDK:
## Problem

`daemon_alive()`, `ensure_daemon()`, `restart_daemon()` (all in `admin.py`), and `daemon.py`'s `serve()` and `already_running()` all call `socket.socket(socket.AF_UNIX, ...)` directly. `AF_UNIX` does not exist on this Windows Python build (`AttributeError: module 'socket' has no attribute 'AF_UNIX'`), so every one of those call sites crashes before a connection is ever attempted.

`helpers.py`'s `_send()` — the actual client RPC call every `cdp()` invocation goes through — has the same issue, so this isn't limited to health checks; the whole daemon protocol is unreachable on this platform.

Reproduced on: Windows 11, Python 3.12.10. `browser-harness --doctor` crashed with a traceback instead of reporting status.

## Fix

New `ipc.py`: one shared module deciding POSIX Unix-socket vs Windows TCP-loopback, so the five call sites share one `connect()`/`listen_kwargs()` pair instead of each duplicating the platform branch.

- POSIX: unchanged behavior — `AF_UNIX` + the existing `/tmp/bu-{name}.sock` path.
- Windows: TCP loopback on `127.0.0.1`, port derived deterministically per daemon `name` via `zlib.crc32` — **not** Python's `hash()`, which is salted per-process (`PYTHONHASHSEED`) and would give the daemon process and a client process different ports for the same name.

`daemon.py`'s `serve()` now picks `asyncio.start_unix_server` or `asyncio.start_server` based on the same `ipc.USE_UNIX_SOCKET` flag.

## Verified live, not just unit-tested

- `browser-harness --doctor` now completes all checks instead of crashing (reports `[FAIL] daemon alive` correctly when none is running, rather than throwing).
- 6 new tests in `test_ipc.py`, including one that opens a real listening socket (whichever family the host actually uses) and round-trips bytes through `ipc.connect()` — not mocked.
- While building this, running `ensure_daemon()` end-to-end surfaced a real `NameError` I'd introduced (a stale variable reference after a rename) that the test suite did not catch — only actually running it did. Fixed, then re-verified live: `ensure_daemon()` now runs **past** the socket layer entirely and fails later at an unrelated, pre-existing issue (`daemon.py`'s `PROFILES` list is macOS/Linux-only paths) — out of scope for this PR, happy to open a separate issue/PR for it if useful.

## Scope

Only the socket family/connection mechanics changed. No behavior change on POSIX (same paths, same `AF_UNIX`, same socket options); this is additive for Windows.

---
## Summary by cubic
Fixes daemon communication on Windows, where `socket.AF_UNIX` doesn't exist and every daemon socket call crashed with `AttributeError` before connecting. Adds `ipc.py` as the single platform decision point, so `browser-harness --doctor` and all client RPCs now complete instead of throwing.

**Bug Fixes**
- All five call sites in `helpers.py`, `admin.py`, and `daemon.py` now use `ipc.connect()` and `ipc.listen_kwargs()` instead of building their own `AF_UNIX` sockets.
- Windows uses TCP loopback on `127.0.0.1` with a port derived from `zlib.crc32` of the daemon name; Python's `hash()` is salted per process, so it can't be used.
- POSIX behavior is unchanged — same Unix socket paths and options.
- Adds `test_ipc.py` with 6 tests, including a live socket round-trip over whichever family the host uses.

<sup>Written for commit 88bdddc32553f9e153a34cb2629e988e65850b0b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/browser-use/browser-harness/pull/811?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LiewSDK:
Closing — opened this against the wrong target by mistake, the fix is being tracked in our own internal system instead. Apologies for the noise.

PR #821, opened 2026-09-19
feat(helpers): support background browser windows

JW-Rami:
## Problem / Context

Agents can already attach to a background tab without changing the visible tab, but they cannot request a separate browser window. This matters when automation shares a logged-in browser while the user keeps browsing. Related public proposals: [#498](https://github.com/browser-use/browser-harness/pull/498) and [#504](https://github.com/browser-use/browser-harness/pull/504).

## Change

- Add `new_tab(url, new_window=False)`.
- With `new_window=True`, create `Target.createTarget` with `background=True` and `newWindow=True`, without activating it.
- Expose the option through `browser_new_tab`.
- Document Brave/macOS setup and shared-profile security.

```mermaid
sequenceDiagram
    Agent->>MCP: browser_new_tab(url, new_window=True)
    MCP->>Helper: new_tab(url, new_window=True)
    Helper->>CDP: Target.createTarget(background=true, newWindow=true)
    CDP-->>User: separate background window
```

## Validation

- `uv run --extra mcp --with pytest python -m pytest tests/unit -q` - 270 passed
- Live macOS Brave smoke: separate window confirmed; frontmost app unchanged; test window closed
- `git diff --check`


JWMatheo:
Must have

